PR #3, opened 2019-11-05
romeo: add robot description

jcarpent:


jcarpent:
The goal is then to rely on this repository for Pinocchio to perform benchmarks, examples, etc.

PR #6, opened 2019-12-03
Migrate universal robot srdf files from hpp-universal-robots.

florent-lamiraux:
The test fails with a mysterious message from boost python.


nim65s:
I don't see how this could make test fail, especially as tests are not yet aware of this new files.

florent-lamiraux:
Test failure was due to a wrong configuration of pinocchio. I did not compile with the python interface.

PR #7, opened 2019-12-05
Minor fine tuning of the project

jcarpent:


wxmerkt:
Is this just cosmetic or does the switching the include order make a functional difference? (Still wrapping my head around jrl-cmake-modules)

jcarpent:
@wxmerkt Just cosmetic change to get first the project message and then the messages coming from CMake itself ;)

Issue #1, opened 2019-08-29
Renaming the project


jcarpent:
@nim65s I think a better name for the project is expected for clarity and self-explainability.
`many-robots-description` may be a candidate.

nim65s:
@cmastalli what do you think about this ?

cmastalli:
To be honest, I don't like neither the keyword "data", instead using description is a good idea seems most people are familiar with this because ROS community. What about just description (without plural)? Instead we could might put the plural in robot, i.e. "robots".

What it doesn't convince me yet is the keyword "many". I remember that @ostasse likes the keyword "example". His point is something like this "I don't want people use these URDFs for our real robots, then using example make clearly state this". @ostasse also mentioned that this is important for our partners (PAL robotics).

So probably we could rename the repository to something like: "example-robots-description" (or shorter "erobots-description"). Personally, I like the second one we can always add a good self-explanatory description of this package.

nim65s:
"erobots" could be understood as "e-robot", which doesn't mean much and would therefore make a good name for science fiction films, or "ero-bots", which mean something else we don't want to see here…

The contraction "e(xample-)robots" is not clear for me.

example-robots-descriptions is fine for me. The 2 plurals can explain both that we have multiple robots and that we have multiple descriptions for some of them, and could replace the "many" suggested by @jcarpent .

nim65s:
but another version of example-robot(s)-description(s) is also good for me

jcarpent:
robot-description-collection

nim65s:
I agree with @olivier-stasse, it has to be clear that these must not be used on real robots, and they can only serve in examples, unit-tests or benchmarks. Maybe "example" is not the best word for that, but this idea has to be in the title

jcarpent:
Weird naming, but lets keep it as people have become familiar to it.

PR #8, opened 2019-12-11
Enhance ur_description

florent-lamiraux:
1. Use meshes from example-robot-data,
2. Remove light from mesh models.


jmirabel:
A remark for later: the lights of DAE files can be turned off in gepetto-viewer.

robots/ur_description/urdf/ur10_joint_limited_robot.urdf
```diff
@@ -32,15 +32,15 @@
   <link name="base_link">
     <visual>
       <geometry>
-        <mesh filename="package://ur_description/meshes/ur10/visual/base.dae"/>
+        <mesh filename="package://example-robot-data/robots/ur_description/meshes/ur10/visual/base.dae"/>
```
nim65s: I already think about that, and changed my mind: https://gepgitlab.laas.fr/gepetto/example-robot-data/merge_requests/29/diffs

Maybe we should discuss a bit about that.

jcarpent: Yes, this will impact everybody (Pinocchio, crocodyl). I think we should move back, otherwise the policy is not the same for all the robots inside example-robot-data!

jcarpent: @nim65s Do you have in mind to revert it?

nim65s: I think we should revert it, yes. 
@florent-lamiraux @jmirabel do you want to add anything ?

florent-lamiraux:
I do not understand the meaning of this package. @nim65s added a dependency to it in hpp-universal-robot. To clean up HPP building, I naively removed dependency to ur_description and transfered useful files (meshes and srdf) to this package.

Please explain the rationale.

robots/ur_description/urdf/ur10_joint_limited_robot.urdf
wxmerkt: It is an accurate change in the meaning of the package - e.g. meshes being found in RViz (path relative to the package.xml). I think these changes could make sense. **However, I think we should have more discussion and review before quickly merging and (potentially) breaking such a widely used package - especially with such a generic title.**  @florent-lamiraux @nim65s @jcarpent

nim65s:
The rationale  of this package is to have an easy to use standalone source for robot descriptions, to use in examples, unit tests, tutorials, showcases and teachings.

Then, this package can be used by other projects to avoid the need for complex dependency management, given that ROS is not really good at keeping old robot descriptions in its new binary releases, and that getting them from source might require the compilation or installation of a ton of dependency, just to get static files in the end, for only one little test / example of your main package.


I moved the description from hpp-universal-robots to here because I believe that in HPP, UR is used for this kind of stuff, and that you don't actually have one real robot and need to adapt the urdf / meshes / collisions to your real robot use.

Then, when one wants to compile HPP, it can just grab the static files from here, and examples / unit tests will work,  without the need for xacro, ros-control and gazebo, which are required for https://github.com/ros-industrial/universal_robot

This is why we have the "example" in the package name, but this is actually still in discussions: https://github.com/Gepetto/example-robot-data/issues/1

wxmerkt:
And this is fine - however, the update that @florent-lamiraux did here was to fix the package paths relative to the main package (`example-robot-data`). I assume this is so that the meshes load correctly in RViz or other packages that resolve with respect to the original package (`example-robot-data`) and not the now-not-existing-former-packages (`ur_description`). As thus, I consider this PR an improvement. 

In what way could it break things that would warrant it to be reverted? [I don't think it should, and I think these fixes should be carried through all the other URDFs]

nim65s:
@wxmerkt : my point about opening discussions are:
1. a change like this can break stuff for other people. If this was working with the old path, there are chances that changing those paths will break stuff, so this must not be merged without clear advertisement.
2. this kind of change must be done at once for all robots, in order to keep a consistency in this package.
3. I don't fundamentally disagree with this change, but I think this is a real question about how this package should be used, and we have to discuss that further.

wxmerkt:
I second (1) and (2).

nim65s:
so, from my point of view, about changing those path:
PRO: for use without the provided loaders, we would not need to use stuff like `ROS_PACKAGE_PATH=<prefix>/share/example-robots-data/robots` anymore, which feels a bit hacky anyway.
CON: the closer we stay from the original sources, the easier merges will be when we will want to update the files.

wxmerkt:
This repository provides assets under different paths. As such we need to update the paths across all URDFs and should open an issue to keep track on those follow-ups.

I don't see CON. If hosted under a different package name, this is the price that has to be paid for that decision to use a different package name and structure.

nim65s:
also in CON, I was thinking that it would be easier to provide a drop-in replacement in dependent packages, without having to change anything in those packages, to ease adoption.  But this point is now obsolete, I think.

jmirabel:
I tend to agree with @wxmerkt. I don't understand the benefits of having the URDF and the meshes in different packages, if there are no dependency between them.

@nim65s said:
> The rationale of this package is to have an easy to use standalone source for robot descriptions, to use in examples, unit tests, tutorials, showcases and teachings.

This goes towards this PR. A URDF without the meshes is not standalone.

> this must not be merged without a clear announcement.

Agreed

> this kind of change must be done at once for all robots, in order to keep a consistency in this package.


For the release, yes. On the devel branch, it's fine if it is split in several PRs.

robots/ur_description/urdf/ur10_joint_limited_robot.urdf
cmastalli: When we started the repo, we aimed to do not change URDF files compared with its main sources. The idea was to quickly update example-robot-data sources once we have new changes in the source files.

If we starte to modify URDF files, then we will lose this trackability, and it's a CON. Probably we could figure out a different solution to this PR.

cmastalli: When we started the repo, we aimed to do not change URDF files compared with its main sources. The idea was to quickly update example-robot-data sources once we have new changes in the source files.

If we starte to modify URDF files, then we will lose this trackability, and it's a CON. Probably we could figure out a different solution to this PR.

jcarpent:
> For the release, yes. On the devel branch, it's fine if it is split in several PRs.

It always better to do everything similar in the same row, compared to the current approach, even on the devel branch. It will simplify a lot the tractability (with git blame) and so on.
I would be in favor to do everything now. And to provide the release asap to adopt it w.r.t the existing packages.

cmastalli:
There is wrong team behavior regarding this repository. It is not good if everyone has the right to merge PR and without any feedback. Only @nim65s and @cmastalli have the rights to accept or to reject PR, other members might earn this right. @nim65s I suggest to define rights per packages to avoid this in future.

On the other hand, I was not aware of this repository, we were always working on the gepgitlab. And we never decide to move the development to github, so this is not good neither.

nim65s:
@jmirabel Agreed, but I never said we should provide urdf and meshes in different packages :)

@cmastalli it's OK to break the devel branch, so i don't think we should monitor merge rights for it. I think other packages should either depend on master or on binary releases anyway. 

I think @florent-lamiraux should not have merged this PR in 3 hours, without a review from  the main authors and without understanding the rationale of the project, but this is clearly OK to do mistakes in devel branches, and users of devel branches should be prepared to handle such mistakes. 

To enforce this, we must prevent unexperimented users to switch to devel just because they need a new feature, and the best solution for this is to "release early, release often". Which I think we're doing well here.

What would IMO be a wrong team behavior would be taking those rights away from experienced users we already trust. This would create a barrier for collaboration and adoption of this package. Also we have way too much packages and users to handle a per-package per-user access.  And in this particular case, I don't think @nim65s & @cmastalli make overall less mistakes like this one than @florent-lamiraux ;)


About your other point, yes, the current state is really wrong. Let me explain:
- all projects from the Gepetto namespace are mirrored from gepgitlab to github
- then, for users with access to gepgitlab, I think it is better to work on gepgitlab, as it provides the CI, and ensure that the 2  repositories keep their synchronisation.
- we discussed that at lengths in HPP-Meetings, as usual without ever reaching a conclusion that make everyone really happy.
- everyone but me seems to prefer mainly working with Github, regardless of the advantages of gepgitlab, as soon as a project is public. Maybe they don't hate closed services and Microsoft as much as I do  x)
- My solutions for this are:
    1. force users to use the tool I maintain, and forbid them to use this evil forge
    2. keep gitlab only for private repositories (and forget about all CI, but it looks like this was not a problem for @florent-lamiraux anyway :P)
    3. revert the mirroring direction, make the Gitlab CI work with Github, and clearly state that every public action should be taken on github, and use gitlab as a forge only for private projects.

Currently, I'm in favor of (iii), but this require some work, and will have to wait early 2020. Until then, I don't have a good solution for this current wrong state of things.

cmastalli:
@nim65s I agree that we could break devel, sometimes. However, this decision has to be done by the main developers of the repository, which are the responsible of any action.  And this is due to the fact that a collaborator rarely will solve a broken code, isn't it?

I prefer option iii, too.

nim65s:
The whole point is that sometimes, one unintentionnaly break stuff. We don't have the workforce to double-check every change in every PR. Then we need a stagging git branch to catch those breaks before they reach production. 

We are all humans, and we all make mistakes. If we don't take this as a premise, every workflow we find is doomed.

Anyway, let's go for (iii), and, after discussions with @florent-lamiraux & @jmirabel, we think we can just make another PR to take those path changes in account in other models. We are aware of some CONs, but they don't seem to prevail the PROs.

PR #13, opened 2020-03-31
Iris model

PepMS:
This PR adds the IRIS quadcopter model


PepMS:
Done! :smile:

cmastalli:
thank you for this contribution :)

wxmerkt:
I don't think we should have merged this and should slow down in the future. **The `iris.stl` is 4MB!** And now since it's in the git history without a history rewrite we won't be able to get it out. If there is a nice and textured dae included (which it seems), let's use it. Do we need the `stl` for collision checking or something else?

cmastalli:
@PepMS do we need the `STL` file? could we replace with the `DAE`?

@wxmerkt - if we can get of rid of the `STL` file, then we could still remove it from the history (by making a forced push). Of course, I could do it if @nim65s agrees with the policy.

We haven't deploy this branch, so everything is still possible.

wxmerkt:
I don't mind right now to follow through with this. But usually the moment something is merged we should never force push.

nim65s:
The issue with history rewrites is that users without enough git training will have issues to get back on track, when their `git pull` will not work, while we don't have a proper communication channel to send them a notice about that.

But as this is the devel branch, our policy is to allow mistakes, so I think we should rewrite the history.

I also think we should avoid that as much as possible, and for this, waiting a working day or so before merging a PR would help a lot.

PepMS:
I'm sorry for this. I've tried to reduce the STL file but I didn't manage to. The other two .dae files are the propeller mesh files.

If you do not want to have such big files in the repo I understand, I'll keep them in my fork.

cmastalli:
@wxmerkt the STL file is needed, one possible action is to try to reduce it as much as possible.

I asked @PepMS and he tried to reduce it with meshmixer.

@wxmerkt Do you know an efficient method to reduce this file?

wxmerkt:
Both Blender and Meshlab may help (they have simplification tools). If not directly with the automatic reduce methods, you may be able to remodel parts of it

PepMS:
Meshlab is a nice tool, @wxmerkt ! I do manage to reduce the size of the file. Now it is 1.4MB

Do you find it acceptable?

wxmerkt:
It's still large for such a simple model, but much better.

cmastalli:
> How should I proceed? Should I create a new PR with the new file?

@PepMS send me the file via email. I need to rewrite the history, and you don't have those rights. Note that PR doesn't rewrite the history!

wxmerkt:
Actually instead of rewriting history you may be able to force-push the version before this PR... this way not the entire history is broken.

nim65s:
Rewriting history should not change anything before the updated commits, but yes, re-opening another PR based on the ancient state of the devel would be cleaner.

cmastalli:
@nim65s do you prefer to open a new PR? I don't have any specific preference.

nim65s:
Yes, let's do this :)

@PepMS: can you open a new PR on the devel branch ? It is now at its older state.

PR #14, opened 2020-04-01
Reduced the STL file of IRIS robot

cmastalli:


wxmerkt:
Merging this branch without prior squashing will not resolve the issue. It still contains the big and the small mesh. We need to squash this to remove the large file from history.

nim65s:
Let's "squash & merge", then :)

Issue #17, opened 2020-04-02
UR robot doesn't work with latest release


cmastalli:
This is the code
```python
import example_robot_data
model = example_robot_data.loadUR()
```
and it appears
```bash
/usr/local/lib/python2.7/dist-packages/pinocchio/shortcuts.py in buildModelsFromUrdf(filename, package_dirs, root_joint, verbose, meshLoader, geometry_types)
     43     for geometry_type in geometry_types:
     44         if meshLoader is None or (not WITH_HPP_FCL and not WITH_HPP_FCL_BINDINGS):
---> 45             geom_model = pin.buildGeomFromUrdf(model, filename, geometry_type, package_dirs)
     46         else:
     47             geom_model = pin.buildGeomFromUrdf(model, filename, geometry_type, package_dirs, meshLoader)

ValueError: Mesh package://example-robot-data/robots/ur_description/meshes/ur5/collision/base.stl could not be found.
```

I don't know what it is the source of the problem, maybe pinocchio?
This issue was reported by @mfocchi

nim65s:
those `package://` stuff are usually loaded by resource-retriever, by looking at ROS_PACKAGE_PATH, or CMAKE_PREFIX_PATH.

I think this `loadUR()` method is ran in unit tests, and [unit tests are passing](https://gepgitlab.laas.fr/gepetto/example-robot-data/pipelines), so I guess this is just about the configuration on your side.

But yes, we should probably advertise that we have a dependency on something able to understand thesepackage:// url.

cmastalli:
This was not necessary before, indeed it is not required for the other robots.

It should fix the issue by returning back the old way of including the meshes file, i.e.
```bash
        <mesh filename="package://ur_description/meshes/ur5/visual/base.dae"/>
```

There is a reason to this change, could we return it back?

---

I really don't like to depend on ROS setup :(

cmastalli:
This was not necessary before, indeed it is not required for the other robots.

It should fix the issue by returning back the old way of including the meshes file, i.e.
```bash
        <mesh filename="package://ur_description/meshes/ur5/visual/base.dae"/>
```

There is a reason to this change, could we return it back?

---

I really don't like to depend on ROS setup :(

wxmerkt:
Pinocchio recently made changes to the URDF loaders. Perhaps it's related? It appears unlikely the issue is in example-robot-data as there haven't been changes to the URDFs themselves?!

cmastalli:
[...]

PR #18, opened 2020-04-03
Fix issue in ur robot #17

cmastalli:
This PR reverts commit 2545adcf0bbecd43ed40d6be03820cb53f3a16d7.

### DIFF
--- a/robots/ur_description/urdf/ur10_joint_limited_robot.urdf
+++ b/robots/ur_description/urdf/ur10_joint_limited_robot.urdf
@@ -32,7 +32,7 @@
   <link name="base_link">
     <visual>
       <geometry>
-        <mesh filename="package://example-robot-data/robots/ur_description/meshes/ur10/visual/base.dae"/>
+        <mesh filename="package://ur_description/meshes/ur10/visual/base.dae"/>
       </geometry>
       <material name="LightGrey">
         <color rgba="0.7 0.7 0.7 1.0"/>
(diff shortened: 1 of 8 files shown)


nim65s:
We already had tons of discussions in #8 so let's wait for others to comment here.

wxmerkt:
No, this is incorrect and will break things when resolving these URDFs via rospy or rospack, or in RViz. 

Background: We do not provide `ur_description`. The package is called `example-robot-data`, and as thus needs to be resolved with respect to the origin of this package. As thus, the fix in #8 was correct. If this causes an issue with Pinocchio or other path resolvers, we should look into fixing it there. Accepting this PR would break the URDFs or require additional packages to be installed (`ur_description` in this case) - making the point of this package kind of moot.

cmastalli:
> No, this is incorrect and will break things when resolving these URDFs via rospy or rospack, or in RViz.
> 
> Background: We do not provide `ur_description`. The package is called `example-robot-data`, and as thus needs to be resolved with respect to the origin of this package. As thus, the fix in #8 was correct. If this causes an issue with Pinocchio or other path resolvers, we should look into fixing it there. Accepting this PR would break the URDFs or require additional packages to be installed (`ur_description` in this case) - making the point of this package kind of moot.

If we reject this PR, then we need to adopt the new policy for all repositories. At the time being, UR description is different to the other cases.

It would be great to modify the loader script in such a way that we don't need to define `ROS_PACKAGE_PATH`. This package should not depend on `ROS`, as any software in the planet.

wxmerkt:
Per the URDF convention, the path should be respective to the root of the package it is referring in `package://`. As the `robots/ROBOTNAME` folders are not the source of a separate package (e.g. there would have to be a `package.xml` specifying the name as `ur_description`), the root is the repository root. And as thus the UR URDF is correct.

Which loader gets tripped up by this and would requite the additional variable set?

cmastalli:
I will close this PR as the proposed solution seems to be not the standard one.

I have created an issue for this topic #19.

Issue #19, opened 2020-04-03
Redefine the mesh path for all robots to match URDF convention


cmastalli:
This task was partially done for the UR robot. We need to
 1. Modify all URDF files to match convention
 2. Add README documentation (optional)

nim65s:
Why would we need any documentation ? The whole point is that everything should work out of the box in all cases, right ?

cmastalli:
> Why would we need any documentation ? The whole point is that everything should work out of the box in all cases, right ?

Yes, I agree. But I am not sure if this possible. I didn't think about it.

jcarpent:
Done.

PR #20, opened 2020-04-23
update solo12

nim65s:
The solo 12 we had was actually a solo 8.

This PR fix this, by importing meshes from https://github.com/open-dynamic-robot-initiative/robot_properties_solo/tree/master/meshes, updating the urdf, and generating some .obj from the .stl when they were missing.


nim65s:
@thomasfla : could you review this ?

robots/solo_description/meshes/obj/solo12/solo_12_base.mtl
```diff
@@ -0,0 +1,9 @@
+# File produced by Open Asset Import Library (http://www.assimp.sf.net)
+# (assimp v5.0.0)
+
+newmtl DefaultMaterial
```
cmastalli, line 4: Why do we need to add all the `.mtl` files?

The looks like control configuration, we are not adding any reference about controllers here.

wxmerkt: `.mtl` is for adding material/color to the meshes when visualising

robots/solo_description/robots/solo12.urdf
```diff
@@ -9,15 +9,15 @@
...
+      <mass value="1.16115091"/>
       <!-- The base is extremely symmetrical. -->
       <inertia ixx="0.00578574" ixy="0.0" ixz="0.0" iyy="0.01938108" iyz="0.0" izz="0.02476124"/>
     </inertial>
     <!-- BASE LINK VISUAL -->
     <visual>
       <origin rpy="0 0 0" xyz="0 0 0"/>
       <geometry>
-        <mesh filename="package://solo_description/meshes/obj/with_foot/solo_body.obj"/>
+        <mesh filename="package://solo_description/meshes/obj/solo12/solo_12_base.obj"/>
```
wxmerkt, line 20: We should update all these paths to `package://example_robot_data/robots/solo_description/...`

nim65s: I agree with this, but this is a separate issue: #19

wxmerkt: Shouldn't we address them directly when adding new URDFs though? When touching Solo, we should in the same step fix its URDFs - don't you agree? Instead of doing a separate action for that immediately after.

nim65s: I am not adding a new URDF here.

I think this might be harder to track changes about one issue if the similar commits are everywhere. 

I also think the state of this repository should be consistent.
The path as they are now are broken for some use case, but working for others. Right now, universal-robot's URDF are not like the other ones, meaning that all tests are passing except those on UR.

If we make this change here, all tests will pass except those on UR & solo 12.

If you suggest to update solo 12, while not updating solo  8, it's going to be non-consistent even in the same robot directory. And if you suggest to update solo 8 and solo  12, this change a bit the scope of this PR.

I think the best option here is to fix #19 first, then rebase this PR on that fix.

wxmerkt: > If you suggest to update solo 12, while not updating solo 8, it's going to be non-consistent even in the same robot directory. And if you suggest to update solo 8 and solo 12, this change a bit the scope of this PR.

My suggestion was to update both Solo URDFs. 

Either way, I think the point is we should fix #19.

nim65s: done :)

robots/solo_description/meshes/obj/solo12/solo_12_base.mtl
nim65s: PDF & png files are already here and not part of this PR, so this is a separate issue.

the `12` option is already in the loader script.

nim65s:
I guess we still need:
1. a review from @thomasfla 
2. an answer to @wxmerkt 

> 2. If the collision objects are .obj, why are we adding the .stl?

For 2., I have no idea. I don't know well the pro/cons of both formats. I would look at their size, the kind of information they embed, and the support in they have in different places. If they are totally redundant, yes, we should only keep the best one.

robots/solo_description/meshes/obj/solo12/solo_12_base.mtl
cmastalli: Thanks :)

wxmerkt:
> For 2., I have no idea. I don't know well the pro/cons of both formats. I would look at their size, the kind of information they embed, and the support in they have in different places. If they are totally redundant, yes, we should only keep the best one.

Usually people have a simplified stl for collision checking and obj or dae for visualisation. I am happy to keep both, if both are used. Here, there are files added that none of the URDFs reference.

nim65s:
Maybe we should wait a bit for https://github.com/open-dynamic-robot-initiative/robot_properties_solo/issues/19https://github.com/open-dynamic-robot-initiative/robot_properties_solo/issues/19

thomasfla:
Looks good to me, we can merge this now and patch it when MPI solves the collision glitch.
Thanks @nim65s !

nim65s:
I started that again.  I added only the new STLs, and I can load the robot, and reproduce @thibnoel last results in https://github.com/open-dynamic-robot-initiative/robot_properties_solo/issues/19

nim65s:
If I run `assimp export solo12_lower_leg_left_side.stl solo12_lower_leg_left_side.obj`, I get:
```
╰─>$ ll solo12_lower_leg_left_side.*
 110K  solo12_lower_leg_left_side.stl
 185K  solo12_lower_leg_left_side.obj
  180  solo12_lower_leg_left_side.mtl
```

So the OBJ are even heavier than the STL, so I don't see the point in using them.

robots/solo_description/robots/solo12.urdf
```diff
@@ -9,15 +9,15 @@
...
+      <mass value="1.16115091"/>
       <!-- The base is extremely symmetrical. -->
       <inertia ixx="0.00578574" ixy="0.0" ixz="0.0" iyy="0.01938108" iyz="0.0" izz="0.02476124"/>
     </inertial>
     <!-- BASE LINK VISUAL -->
     <visual>
       <origin rpy="0 0 0" xyz="0 0 0"/>
       <geometry>
-        <mesh filename="package://solo_description/meshes/obj/with_foot/solo_body.obj"/>
+        <mesh filename="package://example-robot-data/robots/solo_description/meshes/stl/solo12/solo_12_base.stl"/>
```
wxmerkt, line 20: A previous version of this PR had obj including material for the _visual_ meshes. They also appear to still be in the repository as the previous URDF had them. Does this PR change the look to uni-color due to this? I.e. a disimprovement in "looks"?

nim65s: For now, I have a uni-color robot. And according to https://github.com/open-dynamic-robot-initiative/robot_properties_solo/commit/266965a86759f99d196c6f5e4bf6cc61b038e272 I should see something better in gepetto-viewer. But I don't think it was working for me. I'll try again.

cmastalli: @nim65s it seems you're using the STL for visualization. This kind of files doesn't contain color or texture information.

nim65s: It doesn't work. It looks like MTLs are for OBJs. If I add them both, and switch the urdf to obj, I load a bunch of cubes, and the robot inside. 

Plus anyway, the obj are not yet available for solo12, only for solo. If I generate those with assimp, I get the same result, and I don't think I'll get more info.

So, no, this new version of this PR is not a disimprovement in looks, as I don't think I ever see anything else than a uni-color robot.

nim65s: I'v pushed that on the https://github.com/Gepetto/example-robot-data/tree/topic/solo_color branch, if anyone wants to test, but I suggest to keep that as an issue for later.

PR #21, opened 2020-04-26
Fix package paths in URDFs - resolve issues 9,19

### LINKED ISSUE #9: Using meshes provided by example-robot-data
Following recent discussions, I propose to move forward in  the direction of using meshes provided by example-robot-data instead of native description packages.

I will propose a pull request handling robots for which I will have tested that this does not break anything when loading models.

I also propose to revert commit ed85b87 since @cmastalli uses the lighting provided by the models.

### LINKED ISSUE #19: Redefine the mesh path for all robots to match URDF convention

wxmerkt:
This PR resolves #19 with the exception of Solo, which I assume Guilhem will add to #20 (not added here to avoid conflicts).

Resolves #9 
Resolves #19

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package format="2">
+<package format="3">
   <name>example-robot-data</name>
   <version>3.2.0</version>
   <description>Set of robot URDFs for benchmarking and developed examples.</description>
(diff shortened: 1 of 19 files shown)


nim65s:
Thanks for this @wxmerkt !

nim65s:
I double checked it, and it works well both in CI and in my local setup, with a `ROS PACKAGE_PATH` only pointing to the install prefix. Maybe another step would be to add a visual test setting the `ROS_PACKAGE_PATH` to the source dir, to check that we can see the robots before installing the package, but that's another story.

cmastalli:
@nim65s do you think we should describe the need to set up `ROS_PACKAGE_PATH` variable in the README file?

nim65s:
That's a good idea.

wxmerkt:
In a catkin workspace that is what happens when you have a devel or an install workspace automatically. 

If added to the README, then it needs to be explicit that it **only** applies to the vanilla-CMake installation method, otherwise we could see people starting to mess with their ROS_PACKAGE_PATH and end up with hard to debug errors.

nim65s:
Also, on Arch, with https://aur.archlinux.org/packages/example-robot-data, I'm installing it in `/usr`, so it also works with an empty `ROS_PACKAGE_PATH`.

PR #23, opened 2020-04-27
[CMake] pinocchio required only for tests, fix #22

### LINKED ISSUE #22: Distinguish build and runtime dependencies
At present, Pinocchio is required as a build dependency in the CMakeLists. However, I think it should be a runtime dependency only as the Python files are _installed_ and there is no Pinocchio required to compile them.

Changing this avoids having a hard build-time dependency on Pinocchio, and also resolves the circular dependency by having this repository as a submodule in Pinocchio.

I think a viable proposition would be to remove the explicit dependency in the CMakeLists?

nim65s:
In other projects, we the value of BUILD_TESTING, but this means that with default options (just `cmake ..`), if the dependency is not available, the configure fail. On has to use `cmake -DBUILD_TESTING=OFF ..` to let the confgure work (or install the dependency…). 

We could also do that here.

Anyway, my point is that I understand #22 and agree with it, but I would also like to have an error message which is as clear and concise as possible.

Other suggestions ?


unittest/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+IF(NOT pinocchio_FOUND)
+    ADD_TEST(NAME pinocchio-is-required COMMAND ${CMAKE_COMMAND} -E false)
```
wxmerkt, line 2: How about something like this using CMake message directly:

```cmake
message(FATAL_ERROR "Pinocchio is required to build unit tests - please either install Pinocchio or configure without unit tests (-DBUILD_TESTING=OFF)")
```
```

PR #24, opened 2020-04-27
[CMake] Allow to run tests before install

nim65s:
This is a bit ugly, and should probably be done in the jrl-cmakemodule instead, to allow us to write with something like
```python
ADD_PYTHON_UNIT_TEST("py-${TEST}" "unittest/test_${TEST}.py" python 
  ENVIRONMENT "ROS_PACKAGE_PATH=${CMAKE_SOURCE_DIR}/..")
```

Anyway, this doesn't work on Gitlab CI: https://gitlab.laas.fr/gsaurel/example-robot-data/pipelines/9195
Because the project is located in a `example-robot-data-3.2.0` directory, and the URDF is looking for meshes in `example-robot-data/…`…

So I'm just letting this here as a reminder for later.


nim65s:
It works without ROS_PACKAGE_PATH now.

PR #25, opened 2020-04-27
[README] document ROS_PACKAGE_PATH use

nim65s:
While here, and because I have plenty of time because everything is compiling everywhere, I simplified the readme. Here is a rendering: https://github.com/nim65s/example-robot-data/blob/topic/readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,58 @@
-Example robot URDFs
-===============================================
+# Example robot URDFs
 
-## <img align="center" height="20" src="https://i.imgur.com/vAYeCzC.png"/> Introduction
+[![pipeline status](https://gitlab.laas.fr/gepetto/example-robot-data/badges/master/pipeline.svg)](https://gitlab.laas.fr/gepetto/example-robot-data/-/commits/master)
 
-This repository includes a set of robot descriptions that are aimed to be used in benchmarking. These source files do not intend to substitute original their repositories.
+This repository includes a set of robot descriptions that are aimed to be used in benchmarking, unit-tests, teachings,
+tutorials or show-cases. These source files do not intend to substitute their original repositories.
 
+### :construction_worker: Authors
 
-**Authors:**  [Carlos Mastalli](https://cmastalli.github.io/), Guilhem Saurel <br />
-**With additional support from the Gepetto team at LAAS-CNRS and The University of Edinburgh.**
+- [Carlos Mastalli](https://cmastalli.github.io/)
+- Guilhem Saurel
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,52 +1,58 @@
...
-		deb [arch=amd64] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -sc) robotpkg
-		EOF
+   curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key | sudo apt-key add -
+   sudo apt update
+   ```
 
-2. Register the authentication certificate of robotpkg:
+2. installation of example-robot-data and its python utils:
+   ```bash
+   sudo apt install robotpkg-py3\*-example-robot-data
```
cmastalli, line 33: There is a typo here

README.md
```diff
@@ -1,52 +1,58 @@
...
+   ```
 
-		curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key | sudo apt-key add -
+3. If you install this in a non-standard directory and **without catkin**, you will need to set `ROS_PACKAGE_PATH` to
+   your `$CMAKE_INSTALL_PREFIX/share` (`/usr/local/share` by default)
```
wxmerkt, line 37: Does this mean that robotpkg does *not* install to `/opt/openrobots/share` but to `/usr/local/share`?

README.md
nim65s: If I copy-paste that, it works, no matter in you are in 16.04 or 18.04 and have python 3.5 or 3.6.

README.md
nim65s: `/usr/local` is the default `CMAKE_INSTALL_PREFIX. But you are right, we should talk about robotpkg.

nim65s:
@cmastalli : I had a website, but it is currently down. I don't have time for it right now, but I'll setup one some day, and add it here.

README.md
cmastalli: I was thinking in Python 2.5

README.md
wxmerkt: I think we should also add as prominently as a second installation option that simply cloning this into a catkin workspace works just fine, no further configuration needed.

The /usr/local should be option 3 - pure CMake locally.

nim65s: I actually wrote the "install from source" part as you described, and it took a lot of place for only standard boring stuff, so I removed it. What do you think of the current state ? We could add a catkin part, especially if this is really short :)

nim65s:
Hi,

I'd like to make a release with this.
Is https://github.com/nim65s/example-robot-data/blob/topic/readme/README.md OK for you, @cmastalli & @wxmerkt ?

cmastalli:
@nim65s I approve these changes. Thank you for taking care of this :)

wxmerkt:
Good for me, thank you!

PR #28, opened 2020-05-27
Reverted back removation of base_inertia + added tiny inertia in base

cmastalli:
@wxmerkt has modified to make possible simulations with pybullet. However, the previous proposed approach spawn the robot with a different orientation. It seems that pybullet somehow define an initial transformation when this is missed.

This PR reverted back that change because it is needed to define an initial fixed link. Additionally, it includes a tiny inertia in `base` to fix the problem that @wxmerkt handled before.

Note that the initial fixed link is defined as `base` by ANYbotics developers, and then they use an extra link for inertial (called `base_inertia`).

### DIFF
--- a/robots/anymal_b_simple_description/robots/anymal.urdf
+++ b/robots/anymal_b_simple_description/robots/anymal.urdf
@@ -101,25 +101,24 @@
       </geometry>
     </collision>
     <inertial>
-      <origin rpy="0 0 0" xyz="-0.001960558279 -0.001413217745 0.050207125344"/>
-      <mass value="16.793507758"/>
-      <inertia ixx="0.217391101503" ixy="-0.00132873239126" ixz="-0.00228200226173" iyy="0.639432546734" iyz="-0.00138078263145" izz="0.62414077654"/>
+      <mass value="1e-6"/>
+      <inertia ixx="1e-6" ixy="1e-6" ixz="1e-6" iyy="1e-6" iyz="1e-6" izz="1e-6"/>
     </inertial>
   </link>
   <!-- Fixed joint to add dummy inertia link -->
-  <!-- <joint name="base_to_base_inertia" type="fixed">
+  <joint name="base_to_base_inertia" type="fixed">
     <parent link="base"/>
     <child link="base_inertia"/>
     <origin rpy="0 0 0" xyz="0 0 0"/>
-  </joint> -->
+  </joint>
   <!-- Dummy inertia link, because KDL cannot have inertia on the base link -->
-  <!-- <link name="base_inertia">
+  <link name="base_inertia">
     <inertial>
       <origin rpy="0 0 0" xyz="-0.001960558279 -0.001413217745 0.050207125344"/>
       <mass value="16.793507758"/>
       <inertia ixx="0.217391101503" ixy="-0.00132873239126" ixz="-0.00228200226173" iyy="0.639432546734" iyz="-0.00138078263145" izz="0.62414077654"/>
     </inertial>
-  </link> -->
+  </link>
   <!-- Xacro:Properties -->
   <!-- [kg * m^2] -->
   <!-- [A] -->
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -27,27 +27,24 @@ def getModelPath(subpath, printmsg=False):
     raise IOError('%s not found' % subpath)
 
 
-def readParamsFromSrdf(robot, SRDF_PATH, verbose, has_rotor_parameters=True, referencePose='half_sitting'):
-    rmodel = robot.model
-
+def readParamsFromSrdf(model, SRDF_PATH, verbose=False, has_rotor_parameters=True, referencePose='half_sitting'):
     if has_rotor_parameters:
-        pinocchio.loadRotorParameters(rmodel, SRDF_PATH, verbose)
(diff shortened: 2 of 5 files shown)


cmastalli:
@wxmerkt - could you provide feedback to this PR?

wxmerkt:
> However, the previous proposed approach spawn the robot with a different orientation. It seems that pybullet somehow define an initial transformation when this is missed.
> This PR reverted back that change because it is needed to define an initial fixed link.

I cannot fully comprehend your error description and didn't realise a problem with a PyBullet simulation. Could you please explain in more detail or by example?

The tiny inertia is fine as a work-around and should be equivalent.

cmastalli:
> I cannot fully comprehend your error description and didn't realise a problem with a PyBullet simulation. Could you please explain in more detail or by example?

The robot's frame is not aligned with the world frame in the initial posture. I noticed this when I display the Pybullet in Rviz when the grid was not clearly aligned with the horizontal frame defined by the robot's trunk.

It is possible to notice this in Pybullet GUI.

Is this clear?

robots/anymal_b_simple_description/robots/anymal.urdf
```diff
@@ -101,25 +101,24 @@
       </geometry>
     </collision>
     <inertial>
-      <origin rpy="0 0 0" xyz="-0.001960558279 -0.001413217745 0.050207125344"/>
-      <mass value="16.793507758"/>
-      <inertia ixx="0.217391101503" ixy="-0.00132873239126" ixz="-0.00228200226173" iyy="0.639432546734" iyz="-0.00138078263145" izz="0.62414077654"/>
+      <mass value="1e-6"/>
```
jcarpent, line 104: I don’t really like this trick. It also mean that we cannot trust the URDF file anymore when used in external applications. Can you motivate this change?

cmastalli: Gazebo used to require this trick, but to be honest, I don't know exactly the reason. I always assume that it is something related to ODE optimization. @wxmerkt is this still valid? could we remove this trick?

Btw, I have tried in Pybullet, and we don't need this trick.
Indeed, I am confused. I though that Pybullet requires the same trick.
@wxmerkt what was the motivation of your previous change?

wxmerkt: PyBullet assumes a 1kg mass and ixx=iyy=izz=1 inertia if no value is given, so unless you can find a reference of a recent change in PyBullet (pull request), I'd say we do need it

cmastalli: Thanks for the clarification.

Just curiosity, what about Gazebo?

cmastalli:
> This sounds odd given the change. Let me see if I can fully parse this:
> 
>     1. Is your pybullet.loadURDF command with any extra flags?
> 
>     2. You extract the robot frame by using `getBasePositionAndOrientation` and then send it via tf to a RobotModel in Rviz?
> 
>     3. The `base` link/frame is now not aligned with 0,0,0.54/initial-z (or the value given by getBasePositionAndOrientation) in (orientation or position offset)? Or how is not aligned defined?

Yes, I did this. If you zoom in the display in Pybullet, you can also see this offset.

wxmerkt:
Would it be possible to provide a minimal working example or screenshots of before/after?

robots/anymal_b_simple_description/robots/anymal.urdf
wxmerkt: I don't know about Gazebo in its latest versions but I believe it's fine without this work-around now. RBDL would require small inertias as well otherwise it segfaults on loading.

cmastalli:
Screenshot from 2020-05-28 09-35-22
Screenshot from 2020-05-28 09-35-52

robots/anymal_b_simple_description/robots/anymal.urdf
cmastalli: For completeness, ICub and Talos don't have this trick; instead HyQ has it.

I could handle it in this PR. @wxmerkt, @jcarpent  do you agree with it?

wxmerkt:
Thanks, that's very clear. I am a bit confused though: When it looks OK in Rviz, it's tilted in PyBullet and vice versa.

robots/anymal_b_simple_description/robots/anymal.urdf
wxmerkt: I only used HyQ and Anymal during MPC development so that's why the other two dont have it. Happy for you to add it either now or when it's required/validated

wxmerkt:
Just to be clear: How are you loading the URDF in PyBullet? There are flags that can influence the inertia per link and how it represents the tree

robots/anymal_b_simple_description/robots/anymal.urdf
cmastalli: I corrected myself, Talos doesn't need this trick.

I have pushed the modifications of the remaining robots.

cmastalli:
> Just to be clear: How are you loading the URDF in PyBullet? There are flags that can influence the inertia per link and how it represents the tree

```python
import pybullet as pb
import numpy as np
robot_pos0 = np.array([0, 0, 0.64]) # user-defined position (this value is just for explanation purposes)
robot_id = pb.loadURDF(urdf, robot_pos0,  pb.getQuaternionFromEuler([0., 0., 0.]),
                                    flags=pb.URDF_USE_INERTIA_FROM_FILE)

wxmerkt:
^ the flag you are using should work as expected

PR #29, opened 2020-05-30
package.xml: Add catkin/ament_cmake runtime dependency for packaging

wxmerkt:
Required by REP-136

Also adds `build_depend` on Pinocchio as the default for `BUILD_TESTING=ON` - this prevents a configuration error (the error message in `unittest/CMakeLists.txt`) on CI.


nim65s:
Thanks :)

PR #33, opened 2020-06-19
Add missing license file

jcarpent:


cmastalli:
Thank you for adding this missing info

jmirabel:
Isn't it hazardous to provide files copied from other repository under a given license (except if the original license authorizes it) ?

jcarpent:
The license is not for the files distributed, but rather for the Python API. Normally, each modes should come with its own license.

PR #38, opened 2020-06-22
cmake: fix Python install

jcarpent:
Required for Conda


nim65s:
I think we should change `PYTHON_SITELIB` in the jrl-cmakemodule depending on `PROJECT_SITELIB_ROOT`. This would avoid the need for this kind of change here or in any other package.

python/CMakeLists.txt
```diff
@@ -1,12 +1,20 @@
...
 
+IF(PROJECT_SITELIB_ROOT)
+  SET(${PY_NAME}_INSTALL_DIR ${PROJECT_SITELIB_ROOT}/${PYTHON_SITELIB}/${PY_NAME})
+ELSE()
+  SET(${PY_NAME}_INSTALL_DIR ${CMAKE_INSTALL_PREFIX}/${PYTHON_SITELIB}/${PY_NAME})
+ENDIF()
+
 FOREACH(python ${${PROJECT_NAME}_PYTHON_FILES})
-  PYTHON_INSTALL_ON_SITE(${PY_NAME} ${python})
+  INSTALL(FILES
```
nim65s, line 15: `PYTHON_INSTALL_ON_SITE` builds and install. This is useful to check the syntax and the build step, and to get a working python module in the build tree, which can be used by unit tests.

wxmerkt:
I assume the `PYTHON_SITE_LIB` from upstream will still correctly support dist/site packages?

nim65s:
I think that, if defined, `PROJECT_SITELIB_ROOT` should override `PYTHON_SITELIB`.  Otherwise, we should just keep `PYTHON_SITELIB` as it is now.

jcarpent:
`PROJECT_SITELIB` would be ideal, to overwrite the `PYTHON_SITELIB`. Or we just need to allow this overwritting. @nim65s Could you handle it?

PR #41, opened 2020-07-24
[Python] Add (and use) a robot_loader function

nim65s:
To avoid code duplication and ease future robot addition.

While here:
- sync submodule
- use `import pinocchio as pin`
- deprecate `loadTalosLegs()` for the factorized `loadTalos(legs=True)` (and update tests accordingly)
- deprecate `loadTalosArm()` for the factorized `loadTalos(arm=True)` (and update tests accordingly)
- deprecate `loadTalosNoHand()` for the factorized `loadTiago(hand=False)` (and update tests accordingly)
- use cleaner `os.path.join()` for file path manipulation instead of string concatenation
- infer most subpaths ("path / `robots`" in most cases & "path + `srdf`" in all cases)

https://gitlab.laas.fr/gsaurel/example-robot-data/pipelines/10241 (one job is failing because of an on-going CI upgrade)

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -89,10 +89,7 @@ def loadANYmal(withArm=None):
 
 
 def loadTalosArm():
-    URDF_FILENAME = "talos_left_arm.urdf"
-    SRDF_FILENAME = "talos.srdf"
-
-    return robot_loader('talos_data', URDF_FILENAME, SRDF_FILENAME)
+    return robot_loader('talos_data', "talos_left_arm.urdf", "talos.srdf")
 
 
 def loadTalos(legs=False):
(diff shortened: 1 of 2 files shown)


nim65s:
This is a first step of small internal refactors to ease maintenance and new robot additions.

python/example_robot_data/robots_loader.py
```diff
@@ -88,12 +88,8 @@ def loadANYmal(withArm=None):
...
-    return robot_loader('talos_data', "talos_left_arm.urdf", "talos.srdf")
-
-
-def loadTalos(legs=False):
-    URDF_FILENAME = "talos_reduced.urdf"
+def loadTalos(legs=False, arm=False):
+    URDF_FILENAME = "talos_left_arm.urdf" if arm else "talos_reduced.urdf"
     SRDF_FILENAME = "talos.srdf"
 
     robot = robot_loader('talos_data', URDF_FILENAME, SRDF_FILENAME, free_flyer=True)
```
cmastalli, line 95: For the arm case, we should define `free_flyer` as `False`

nim65s: Thanks for your review. This is fixed, but your remark raise the following points:
1. According to `TalosArmTest`, this arm is not supposed to have a free flyer, but according to `TalosArmFloatingTest`, this arm is supposed to have a free flyer.
2. Tests are not testing anything. if I change the RobotTestCase.NQ for one or the other of those tests, the test are still passing. I was going to refactor those anyway.

cmastalli: I would remove the `TalosArmFloatingTest`. The reasons are
 - it is contra-intuitive to have a "floating-base" arm
 - this unit-test doesn't bring anything new

I agree the "tests" are not that good. Nice that you will plan to improve them :)

PR #42, opened 2020-07-27
[solo] disable collision pairs for adjacent links

pFernbach:
Without this changes, the robot is always in self-collision according to hpp-fcl. 

Should I also push this in https://github.com/open-dynamic-robot-initiative/robot_properties_solo ? This repo doesn't have any srdf file.

### DIFF
--- a/robots/solo_description/srdf/solo.srdf
+++ b/robots/solo_description/srdf/solo.srdf
@@ -1,5 +1,5 @@
 <?xml version="1.0" ?>
-<robot name="hyq">
+<robot name="solo">
 
     <!-- left front leg -->
     <group name="lf_leg">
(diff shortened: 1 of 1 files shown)


jmirabel:
Yes, I think it would be good to have the SRDF file in the repo you mentioned. This repo is not supposed to be the main repo for any robot.

cmastalli:
> Should I also push this in https://github.com/open-dynamic-robot-initiative/robot_properties_solo ? This repo doesn't have any srdf file.

I think this should be answered by the main developers / responsables of that repo.

cmastalli:
If there is no anymore to be done, then let's merge it :)

pFernbach:
PR done in the main repository for solo: https://github.com/open-dynamic-robot-initiative/robot_properties_solo/pull/22

nim65s:
Hum, I shouldn't have merged this yet, as we have some inputs on the other PR. We might have to do another PR later, sorry.

PR #44, opened 2020-07-30
[Python] add more paths to getModelPath, fix #37

### LINKED ISSUE #37: Tests fail in `make release` without installation
Tests still fail in `make release` without installation. But this can be fixed later.

_Originally posted by @nim65s in https://github.com/Gepetto/example-robot-data/pull/36#issuecomment-646962947_

nim65s:
and document their purpose


python/example_robot_data/robots_loader.py
```diff
@@ -3,21 +3,23 @@
 from os.path import dirname, exists, join
 
 import numpy as np
-
 import pinocchio as pin
 from pinocchio.robot_wrapper import RobotWrapper
 
 pin.switchToNumpyArray()
```
cmastalli, line 9: This is not connected with your PR, however, we could remove this line.

nim65s: Yes, sorry, my isort seems not to be exactly configured in the same way on my different computers, which sometimes lead to this kind of useless changes. I noticed this one after the PR opening, and was too lazy to remove it, as it really doesn't harm.

PR #45, opened 2020-08-27
add talos full

nim65s:
To avoid a dependency to https://github.com/stack-of-tasks/talos-data/ in many packages.


cmastalli (commented):
Hi @nim65s 

What is the reason to add new mesh files? It seems the `.STL` are copies of the `.stl`.
On the other words, could we use the `.stl` inside `talos_full_v2.urdf`?

Carlos

nim65s:
Thanks @cmastalli, that's right, I fixed that.

cmastalli:
Checking again, I see a mixture of `.stl` and `.STL`, maybe it makes sense to clean up the repo. For instance, we could make all of them `.stl` and modify the urdf respectively.

nim65s:
Yes, this is sad, but I don't see any reason to do this. Everything already works fine this way, and each modification we made at this point will have to be taken in account if we ever want to update anything from the upstream repo.

cmastalli (approved):
I understand the motivation of keeping `stl` and `STL`.
Maybe we should solve this first in the root repo, but we depends upon the agreement with other partners.

Having a single format will avoid duplication.

PR #48, opened 2020-09-15
Fix Hyq srdf (disable collision pairs)

pFernbach:
Disable collision pairs for all the adjacents link. 

Add all the "never" colliding collision pairs. For this last point, I did it symmetrically with respect to what was already done for one leg. But I do not really agree with it: a lot of this pairs are never colliding only if we respect the joints limits specified in the URDF. 

I am not sure about what is the policy for this, for me "never" colliding should only be used when it is physically impossible that both links are colliding.


cmastalli:
I agree with you.

cmastalli (approved):

pFernbach:
Ok, I removed some of this lines then.

nim65s:
This fix some unit tests for me, thanks a lot !

PR #49, opened 2020-09-16
Release/3.6.1

nim65s:
Maintenance release


cmastalli:
@nim65s is this release solving the issue with the cppadcg cmake module?

nim65s:
nope, that's the current devel only.

Issue #52, opened 2020-09-21
Cmake issue when Pinocchio is installed with CppAD support


cmastalli:
I cannot compile this repository if I install Pinocchio with CppAD support (I am using the latest devel branch).
This is the error message when I typed `cmake ../`:

```
CMake Error at /usr/local/lib/cmake/pinocchio/pinocchioConfig.cmake:126 (find_package):
  By not providing "Findcppadcg.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "cppadcg", but
  CMake did not find one.

  Could not find a package configuration file provided by "cppadcg"
  (requested version 2.4.1) with any of the following names:

    cppadcgConfig.cmake
    cppadcg-config.cmake

  Add the installation prefix of "cppadcg" to CMAKE_PREFIX_PATH or set
  "cppadcg_DIR" to a directory containing one of the above files.  If
  "cppadcg" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  cmake/package-config.cmake:84 (find_package)
  CMakeLists.txt:35 (ADD_PROJECT_DEPENDENCY)
```

nim65s:
This is a known issue from jrl-cmakemodules / pinocchio, ref. https://github.com/loco-3d/curves/pull/50.

PR #53, opened 2020-09-21
Added the CppADCoGen cmake path

cmastalli:
This solves #52.


nim65s:
I'm totally against this.

Please read https://github.com/loco-3d/curves/pull/50.

cmastalli:
> I'm totally against this.
> 
> Please read [loco-3d/curves#50](https://github.com/loco-3d/curves/pull/50).

I agree with the statements inside that PR. But inspecting how long it has been there (inside jrl-cmake), I ask myself is we could use this PR for a temporal patch.

If you agree with it, I could add a TODO text. Otherwise, is there another "quick" (and better) solution?

nim65s:
I don't want to manage this.

Do you understand that we need this line in each and every package depending  on pinocchio ? That's close to 100.

Adding a workaround line like that in all our packages is a totally crazy maintenance cost that I'm not willing to pay at all, especially since this will be useful only to people that enable really advanced features and  clearly know how to handle that by themselves while not polluting the main shared repository.

And in that case, It's way easier to edit the installed cmake files from pinocchio instead of the CMakeLists of all the packages you are going to compile.

I'm hugely late for the robotpkg release of the whole SoT, loco-3d and HPP, (in which Carlos you need a crocoddyl release, and you Justin an hpp-fcl update, not mentionning many many other people that have been asking some of the included stuff for literal months), so I'm not going to loose any second for this.

I'm not going to loose time to fight either, so if you want to merge it, go ahead, but that's without me.

cmastalli:
> I don't want to manage this.
> 
> Do you understand that we need this line in each and every package depending on pinocchio ? That's close to 100.
> 
> Adding a workaround line like that in all our packages is a totally crazy maintenance cost that I'm not willing to pay at all, especially since this will be useful only to people that enable really advanced features and clearly know how to handle that by themselves while not polluting the main shared repository.
> 
> And in that case, It's way easier to edit the installed cmake files from pinocchio instead of the CMakeLists of all the packages you are going to compile.
> 
> I'm hugely late for the robotpkg release of the whole SoT, loco-3d and HPP, (in which Carlos you need a crocoddyl release, and you Justin an hpp-fcl update, not mentionning many many other people that have been asking some of the included stuff for literal months), so I'm not going to loose any second for this.
> 
> I'm not going to loose time to fight either, so if you want to merge it, go ahead, but that's without me.

@nim65s I perfectly understand your point, I wouldn't merge it if you are not happy with this solution.
We could leave this branch for anyone that needs a quick solution.

PR #55, opened 2020-10-19
Remove lights from Anymal robot

jcarpent:


nim65s:
I don't really get the rationale, but #8 removed the lights from ur and was accepted, so I guess we should also accept this here.

nim65s (approved):

jcarpent:
The idea is to remove lights that are not welcome when one's perform rendering of the scene
There is no reason to have lights in DAE files, except the ones coming from the internal illumination of the scene.

cmastalli:
How does the visualization change in GVr?
Could you share a pic?

If I remember correctly, it was badly displayed in GV without the lights.

jmirabel:
For reference, there is a `gui.removeLightSources` in GV that remove all lights from a node.

jcarpent:
> How does the visualization change in GV?
> Could you share a pic?
> 
> If I remember correctly, it was badly displayed in GV without the lights.

If you want a better illumination in GV, you need to add an omnidirectional light on top of the scene. But the lights should not be embedded inside the meshes. Otherwise, when the robot moves, the lighting changes too.

jcarpent:
> For reference, there is a `gui.removeLightSources` in GV that removes all lights from a node.

Thanks for the hint.

PR #56, opened 2020-10-23
python: allow loading of multiple robots

nim65s:
Hi,

The first commit is because I saw in our code base that people are using helpers to load our robots, but then need urdf / srdf paths for something else, and have to recompute that by themselves.

The second commit was a request from @thomasfla

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -470,24 +470,28 @@ ROBOTS = {
 }
 
 
-def loader(name, display=False):
+def loader(name, display=False, rootNodeName=''):
     """Load a robot by its name, and optionnaly display it in a viewer."""
     if name not in ROBOTS:
         robots = ", ".join(sorted(ROBOTS.keys()))
         raise ValueError("Robot '%s' not found. Possible values are %s" % (name, robots))
     inst = ROBOTS[name]()
     if display:
-        inst.robot.initViewer(loadModel=True)
-        inst.robot.display(inst.q0)
+        if rootNodeName:
+            inst.robot.initViewer()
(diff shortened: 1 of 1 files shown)


cmastalli (commented):
I would suggest to have a single function for retrieving the paths, rather than appending them inside `robot`.

Indeed, it would be better to do the same for `q0`.

nim65s:
If we try to do this with a simple user-facing API, I think we only end up with a RobotWrapperWrapper, which proxy RobotWrapper plus add `q0` and 3 paths. Is this what we want ?

nim65s:
Another option would be to expose those things directly in pinocchio's RobotWrapper. For me, the main issue is that `q0` collides with something already defined differently in pinocchio: 

https://github.com/stack-of-tasks/pinocchio/blob/17c479e2fe91905122713643f489feb8ddf135a3/bindings/python/pinocchio/robot_wrapper.py#L33

proyan:
Could we merge this? Who has write access on this repo?

proyan (approved):

cmastalli:
> Could we merge this? Who has write access on this repo?

The current solution appends the path into the robotwrapper class.
It would be ideal to do not do it, it also requires the user to learn that we pass a modified version of the robotwrapper.

What I propose is to create a dedicated function to retrieve the paths, and eventually another one to get the default state.

nim65s:
@cmastalli : We can't have a "simple" function, can we ? Could you outline the prototype of the function you are thinking about ?

cmastalli:
> @cmastalli : We can't have a "simple" function, can we ? Could you outline the prototype of the function you are thinking about ?

What about
 - `getUrdfPath('robot')`
 - `getSrdfPath('robot')`
 - `getDefaultConfiguration(`robot`)`
?

(or something like that)

PR #58, opened 2020-11-20
RobotLoader

### LINKED ISSUE #57: Uniform API of the loaders
Hi guys,

The API of the robot loaders is too confusing. For example, to load solo, the API: `loadSolo()`, and to load solo12, API: `loadSolo(False)`

The reason we have the `load` functions, is to be simplify the loading of the urdfs/srdfs by just one single command. In order to do that, the `load<robotName>` the `robotName` here should be clear enough.

However, what I often have to do is to look at the implementation inside the `load...` function to see what is being loaded with which argument. Or I have to look at the arguments to predict what is being loaded where.

For example, consider `loadTalos(legs=False, arm=False, full=False, box=False)`
Here, we have 2**4=16 possible arguments. However, only some of these arguments are valid, and the rest are non-sense. If I want to load the legs of Talos, I need to predict which specific argument to give. Or I need to understand that full=True doesn't equate to arm=True  + legs=True. Finally, I end up looking at the implementation each time to make sure that I'm loading the right thing.

If eventually someone has to go through the implementation, he/she would directly look at the implementation of RobotWrapper and then we don't need these _easy to use_ loaders anymore.

What was the reason to deprecate loadTalosArms or loadTalosLegs?

nim65s:
Hi,

This is a suggestion following discussions in #56.

With this, the main entry point of the public API becomes either:

```python
robot = load('talos')
```

or, if someone wants more that just the robot wrapper,

```python
robot, q0, urdf, srdf = load_full('talos')
```

This suggestion:
- make the code more generic
- avoid multiple calls to `getModelPath()` (which would be necessary for https://github.com/Gepetto/example-robot-data/pull/56#issuecomment-717980586)
- fix #57 
- is open for comments

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -9,6 +9,10 @@ from pinocchio.robot_wrapper import RobotWrapper
 pin.switchToNumpyArray()
 
 
+def _depr_msg(deprecated, key):
+    return "`%s` is deprecated. Please use `load('%s')`" % (deprecated, key)
+
+
 def getModelPath(subpath, printmsg=False):
     source = dirname(dirname(dirname(__file__)))  # top level source directory
     paths = [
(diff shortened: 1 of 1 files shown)


cmastalli (approved):
I am happy with the new API.
Thanks a lot for improving it :)

PR #62, opened 2020-12-04
report warnings to end users, fix #60

### LINKED ISSUE #60: API deprecation method doesn't report message with "python ..."
I can only see the messages when we run `ipython ...`.
I think fixing this issue is important to let users know the deprecation.

nim65s:
`DeprecationWarning` are intended for other python developers. But this package aims to be used by inexperienced end users, therefore we should ensure that they always see our warnings about deprecation.


cmastalli (approved):
Thanks for a very quick fix :)

PR #64, opened 2021-01-08
Fix Anymal SRDFs

wxmerkt:
Fixes errors in subgroup naming and references to virtual joints as identified by the `srdfdom` parser. It also fixes a typo in the Kinova URDF and adds additional planning groups.


cmastalli (commented):
It looks good for me, but I would like to double confirm this with @pFernbach.
Pierre is using some of this details in his HPP stuffs.

nim65s (approved):

pFernbach:
I am not using the group tag so it's safe for me, thanks for checking.

cmastalli:
Thanks @wxmerkt ;)

PR #66, opened 2021-02-11
Define unknown materials to hide URDF parser warnings

wxmerkt:
While debugging Crocoddyl unit tests, we often see a ton of warnings about materials that can't be found which makes it hard to identify the error messages. Defining the missing materials silences all warnings and we can focus on the error messages without scrolling :sunglasses:


jcarpent (approved):

nim65s (approved):
That's really helpful, thanks :)
we are going one step further away from the original files, but the diff is trivial, so I'll be in favor of merging this.

cmastalli (approved):

PR #68, opened 2021-03-03
[talos_reduced] Fix inertia and mass to zero for optical links.

### LINKED ISSUE #67: Talos not correct for pybullet - No link should be left without inertia
In pybullet, links should have an inertia and a mass specified even if this is zero.
Otherwise the default mass value is 1 Kg.
See https://github.com/robotology/icub-models/issues/12

A PR follows fixing this issue.

olivier-stasse:
Fix #67


cmastalli (approved):

nim65s:
Thanks for this !

I think it could be nice to have an unit test that load all robots on pybullet and checks if we get any wrong inertia. I can handle this if you don't have time for it.

olivier-stasse:
@nim65s that would be awesome. A simple test would be loading the file with both pinocchio and pybullet. Both should provide the same mass. Otherwise the inertia parameter is not set for a link.

PR #72, opened 2021-03-29
getModelPath: add path for installed wheels

nim65s:
This will allow us to provide PyPI support.


jcarpent (commented):
Why openrobot should appear here?

nim65s:
I misunderstood how data extraction from wheel works. This is useless.

PR #73, opened 2021-04-07
Removing duplication of q0 in RobotLoader, instead using the one introduced in RobotWrapper

cmastalli:
The `q0` duplication confused me, as I was not sure why the `RobotWrapper.q0` was not filled during calling `load` or `load_full` functions.

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -66,8 +66,8 @@ class RobotLoader(object):
 
         if self.srdf_filename:
             self.srdf_path = join(self.model_path, self.path, self.srdf_subpath, self.srdf_filename)
-            self.robot.q0 = readParamsFromSrdf(self.robot.model, self.srdf_path, self.verbose, self.has_rotor_parameters,
-                                         self.ref_posture)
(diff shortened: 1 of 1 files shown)


hrp2-14:

jcarpent (approved):

nim65s (changes requested):
You are right, we should do this. But fixing this that way might break working code in the wild.

nim65s:
I suggest to add a proxy, eg.:
```python
class RobotLoader:
    @property
    def q0(self):
        return self.robot.q0
```

What do you think ?

cmastalli:
> I suggest to add a proxy, eg.:
> 
> ```python
> class RobotLoader:
>     @property
>     def q0(self):
>         return self.robot.q0
> ```
> 
> What do you think ?

Yes, it sounds a good idea for supporting `RobotLoader.q0`. However, I would include also a deprecating code of this property, do you agree?

nim65s:
yes

cmastalli:
> yes

Done! @nim65s

nim65s (approved):

Issue #75, opened 2021-04-16
parsing error while loading universal robot urdf/dae files


ntorresalberto:
Hi,
From what I read [here](https://github.com/stack-of-tasks/pinocchio/issues/838) it could be related to the hpp-fcl library.

Basically I'm trying this:

```python
import example_robot_data
robot = example_robot_data.load('ur10')
```

```bash
$ python -i mwe_load_ur10.py 
Traceback (most recent call last):
  File "mwe_load_ur10.py", line 2, in <module>
    robot = example_robot_data.load('ur10')
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/example_robot_data/robots_loader.py", line 506, in load
    return loader(name, display, rootNodeName).robot
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/example_robot_data/robots_loader.py", line 493, in loader
    inst = ROBOTS[name]()
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/example_robot_data/robots_loader.py", line 64, in __init__
    self.robot = RobotWrapper.BuildFromURDF(self.urdf_path, [join(self.model_path, '../..')],
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/pinocchio/robot_wrapper.py", line 17, in BuildFromURDF
    robot.initFromURDF(filename, package_dirs, root_joint, verbose, meshLoader)
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/pinocchio/robot_wrapper.py", line 21, in initFromURDF
    model, collision_model, visual_model = buildModelsFromUrdf(filename, package_dirs, root_joint, verbose, meshLoader)
  File "/home/user/.conda/envs/pinocchio/lib/python3.8/site-packages/pinocchio/shortcuts.py", line 45, in buildModelsFromUrdf
    geom_model = pin.buildGeomFromUrdf(model, filename, geometry_type, package_dirs)
ValueError: Mesh index: 0
Face index: 0
```

I can reproduce the same error while using pinocchio with [ros-industrial](https://github.com/ros-industrial/universal_robot) while trying to `buildFromUrdf` to load ur3, ur3e, ur5e, ur10, ur10e.

Some info about my setup: I'm running Fedora 32 and all libraries were installed from conda-forge.

jcarpent:
This issue is due to Assimp in fact. I will try to fix it by regenerating correctly the mesh.

[...]

PR #76, opened 2021-04-16
Unit test for all UR robots

cmastalli:
This is related to issue #75.
Additionally, it is expected that the unittests will not pass, as we need a new release after merging the following PR:
   https://github.com/humanoid-path-planner/hpp-fcl/pull/219


cmastalli:
@nim65s  I see that hpp-fcl v1.7.2 has not been deployed in a few distros.
This is the reason why some of then are failing.
Furthermore, I noted that these CI of these distro are allowed to fail now.
Could you let me know if the policy has changed? Are we still supporting those distros?

cmastalli:
I will merge this PR as it adds important unit-tests.

PR #78, opened 2021-04-29
Fix Anymal-Kinova SRDF collision matrix

wxmerkt:


nim65s (approved):
I'll trust you on this one :)

PR #79, opened 2021-05-04
Define q0 as neutral if the robot doesn't have a SRDF

cmastalli:
Without this rule, I triggered an error in some hector-based examples defined in Crocoddyl.


hrp2-14:

PR #81, opened 2021-05-17
Add baxter robot

nim65s:
This is useful as  [it is no longer available in ROS since indigo](http://wiki.ros.org/baxter_description).


wxmerkt (commented):
I don't think the STL files are used in the URDF - if not, can we exclude them?

nim65s:
Well spotted, thanks !
This saved 13Mo :)

jcarpent (approved):
Nice.

PR #85, opened 2021-05-28
baxter: add grippers

nim65s:
Baxter is commonly used to showcase manipulation tasks. But I forgot to add grippers to the previous version… Here is a fix.
before:
before
after:
after


hrp2-14:

PR #89, opened 2021-06-11
Add all collision pairs and remove disabled pairs from SRDF

wxmerkt:
So far, the RobotLoader does not enable collision pairs by default. Since we have the SRDFs that should disable invalid pairings, we can enable all pairings and then use the SRDF to disable correspondingly.


hrp2-14:

nim65s (approved):
Thanks :)

cmastalli (approved):

PR #93, opened 2021-06-15
Rename "Solo" to "Solo8". Fixes #92

### LINKED ISSUE #92: Rename Solo to Solo8v2
There is a slight naming error right now in the package. When calling `load("solo")` it creates a Solo8. Note however, that "Solo" is the name of the quadruped platform as such, not of any specific robot.

Therefore, it would be better to have `load("solo8")`. Then, there are two versions of solo8: One with TI cards (called `solo8v1`) and one with udrivers (called `solo8v2`). They have different weights. It looks like the one in the repo is from the `solo8v2` model.

Therefore, would it make sense to make `load("solo")` become `load("solo8v2")`?

jviereck:


python/example_robot_data/robots_loader.py
```diff
@@ -258,21 +258,21 @@ def loadHyQ():
...
 
-class Solo12Loader(SoloLoader):
+class Solo12Loader(Solo8Loader):
     urdf_filename = "solo12.urdf"
 
 
-def loadSolo(solo=True):
-    warnings.warn(_depr_msg('loadSolo()', 'solo'), FutureWarning, 2)
-    loader = SoloLoader if solo else Solo12Loader
+def loadSolo(solo8=True):
```
nim65s, line 273: I think this one should keep `solo=True`, because if someone has an old code base with `robot = loadSolo(solo=True)`, we want to load `solo8`and display the warning.

python/example_robot_data/robots_loader.py
```diff
@@ -258,21 +258,21 @@ def loadHyQ():
     return HyQLoader().robot
 
 
-class SoloLoader(RobotLoader):
+class Solo8Loader(RobotLoader):
```
nim65s, line 261: The main `ROBOTS` dict still contain `SoloLoader`, and I think it would be a good thing to add it for a deprecation message, eg.:

```python
class SoloLoader(Solo8Loader):
    def __init__(self, *args, **kwargs):
        warnings.warn('"solo" is deprecated, please try to load "solo8" or "solo12"')
        return super(SoloLoader, self).__init__(*args, **kwargs)
```


(and keep the change you made to the new class)

nim65s (changes requested):
I think you'll also want to add `'solo8': Solo8Loader` in the `ROBOTS` dict

jviereck:
Thanks for the precise review @nim65s . I will work in the comments later today.

python/example_robot_data/robots_loader.py
jviereck: Agreed, thanks for proposing this. I switched it back to `solo`.

nim65s (approved):
This is good for me, thanks !

jviereck:
Is there anything missing for merging this PR?

cmastalli:
I have merged it :)
Thanks for the contribution.

PR #100, opened 2021-07-27
Add cassie robot in example-robot-data

proyan:
This draft PR introduces the cassie_description, which is taken from https://github.com/proyan/cassie-gazebo-sim.
The model introduced is `sdf` format. I use the `cassie_v2.sdf` to produce examples for pinocchio3, and which we reproduced in the paper. The python bindings for the eventual `loadCassie` would thus depend on the pinocchio version being installed.

I open this PR to discuss how to 
1) best handle the multiple other cassie models available (https://github.com/DAIRLab/dairlib/issues/17), and what to introduce and not.
2) How to handle the API for different sdf models as well. I re-wrote `cassie_v2` from the original sdf file which when parsed with pinocchio makes sure that the constraint appears on the ball joint. We should make it the default model loaded, but it isn't the standard model available from Agility.
3) Confirm Licenses when copying the files


proyan:
@nim65s This PR is okay for review

python/example_robot_data/robots_loader.py
```diff
@@ -239,6 +239,50 @@ def loadTalosArm():
...
+
+                # Recreate collision data since the collision pairs changed
+                self.robot.collision_data = self.robot.collision_model.createData()
+        else:
+            self.srdf_path = None
+
+        if self.free_flyer:
+            self.addFreeFlyerJointLimits()
+
+def loadCassie():
```
nim65s, line 281: I don't see the point in creating this function, as it is already deprecated

python/example_robot_data/robots_loader.py
```diff
@@ -531,6 +575,10 @@ def loadIris():
 }
 
 
+if version.parse(pin.__version__) >= version.parse("2.9.2"):
+    ROBOTS['cassie']= CassieLoader
```
nim65s, line 579: I don't see the need for the conditional

python/example_robot_data/robots_loader.py
```diff
@@ -239,6 +239,50 @@ def loadTalosArm():
...
+    verbose=False
+
+    def __init__(self):
+        sdf_path = join(self.path, self.sdf_subpath, self.sdf_filename)
+        self.model_path = getModelPath(sdf_path, self.verbose)
+        self.sdf_path = join(self.model_path, sdf_path)
+        self.robot = RobotWrapper.BuildFromSDF(self.sdf_path, [join(self.model_path, '../..')],
+                                                pin.JointModelFreeFlyer() if self.free_flyer else None)
+        q0 = pin.neutral(self.robot.model)
+        q0[2] = 1.104;    q0[9] = 0.298;    q0[20] = 0.298;    q0[17] = -1.398
```
nim65s, line 256: I thought you already modified the robot. Why not updating it with the data you need ?
I think robot description data belong to urdf / srdf / sdf files, while python files should only be there to reduce boilerplate

Also, this is breaking the format CI with:
`./python/example_robot_data/robots_loader.py:256:22: E702 multiple statements on one line (semicolon)`

nim65s (commented):
Hi @proyan,

I made a proposition to ease this in #101. The idea is to be more generic
If we follow it, I think we don't need any pinocchio version check here.

We will need one to activate or not a unit test for cassie.
And at that point, I would prefer a comprehension list over adding a new dependency, eg. 
```python
tuple(int(v) for v in pinocchio.__version__.split('.')) >= (2, 9, 2)
```

Currently, the CI is failing because of this:
```
    from .robots_loader import (ROBOTS, getModelPath, load, load_full, loadANYmal, loadDoublePendulum, loadHector, loadHyQ,
  File "/root/robotpkg/robots/py-example-robot-data/work/example-robot-data-3.10.0/python/example_robot_data/robots_loader.py", line 8, in <module>
    from packaging import version
```

cmastalli:
>     1. best handle the multiple other cassie models available (

We could have different Cassie versions, for instance,
```python
import example_robot_data
cassie = example_robot_data.load("cassie") # includes gearbox and planar linkage
cassie = example_robot_data.load("cassie_without_mechanism") # do not include gearbox and planar linkage
```

I believe these two cases might be useful for many of us. But, of course, we might create a more elaborated list of Cassie models.

python/example_robot_data/robots_loader.py
proyan: I modified the kinematic tree in the original robot sdf, but not it's initial configuration. I agree initial configuration belongs to `srdf`, but I don't think I would have enough time to create an srdf with cassie. If you think there is a stop-gap measure, we can use that, or keep this PR open till I can push an `srdf`

python/example_robot_data/robots_loader.py
proyan: I was tottering on whether to include a version check or not, but I agree it doesn't add much value

proyan: I've removed the commit

proyan:
> I made a proposition to ease this in #101. The idea is to be more generic
If we follow it, I think we don't need any pinocchio version check here.

If you merge #101 before, then the CassieLoader could be simplified as well. However, I don't see a good way to handle multiple versions of the same robot. However, It might be useful if we can take the urdf/sdf filename inside the arguments list, since that way the custom naming could be avoided. I'm not sure how it might be best handled with the existing API. 
I'll wait for #101 before modifying this PR

proyan:
This PR is ready for review @nim65s

unittest/test_load.py
```diff
@@ -42,7 +42,10 @@ def test_anymal_kinova(self):
 
     def test_baxter(self):
         self.check('baxter', 19, 19)
-
+        
+    # def test_cassie(self):
+    #     self.check('cassie', 29, 28)
```
nim65s, line 47:
```suggestion
    def test_cassie(self):
        try:
            self.check('cassie', 29, 28)
        except ImportError:
            self.assertGreaterEqual(int(pin.__version__.split('.')[0]), 3)
```

nim65s (commented):
Please fix flake8:
```
./python/example_robot_data/robots_loader.py:271:1: E302 expected 2 blank lines, found 1
./unittest/test_load.py:45:1: W293 blank line contains whitespace
./unittest/test_load.py:48:1: W293 blank line contains whitespace
```
plus I have a suggestion, but it's open to discussions

proyan:
> Please fix flake8:
> 
> ```
> ./python/example_robot_data/robots_loader.py:271:1: E302 expected 2 blank lines, found 1
> ./unittest/test_load.py:45:1: W293 blank line contains whitespace
> ./unittest/test_load.py:48:1: W293 blank line contains whitespace
> ```

What version of flake8 do you use for ci? I get many more suggested fixes than the above 2

nim65s:
That's from https://github.com/Gepetto/linters : you can run it with `docker run --rm -v $PWD:/root/src -it gepetto/linters`. It's the latest version, with the following configuration: https://github.com/Gepetto/linters/blob/master/setup.cfg#L1

nim65s:
Oh, and I forgot that we need a `import pinocchio as pin` in the test file

nim65s (approved):

unittest/test_load.py
proyan: I think we needed assertLess here

nim65s: Yes, I pushed that in cb9e824

proyan:
Could we please merge this, and make a release?

PR #101, opened 2021-07-28
[loader] SDF support

nim65s:
This is a proposition to add support for SDF models in pinocchio >= 2.9.2 in `RobotLoader`

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -61,6 +61,8 @@ class RobotLoader(object):
 
     def __init__(self):
         if self.urdf_filename:
+            if self.sdf_filename:
+                raise AttributeError("Please choose between URDF *or* SDF")
             df_path = join(self.path, self.urdf_subpath, self.urdf_filename)
             builder = RobotWrapper.BuildFromURDF
         else:
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -50,19 +50,29 @@ class RobotLoader(object):
     path = ''
     urdf_filename = ''
     srdf_filename = ''
+    sdf_filename = ''
     urdf_subpath = 'robots'
     srdf_subpath = 'srdf'
+    sdf_subpath = ''
     ref_posture = 'half_sitting'
     has_rotor_parameters = False
     free_flyer = False
     verbose = False
 
     def __init__(self):
-        urdf_path = join(self.path, self.urdf_subpath, self.urdf_filename)
-        self.model_path = getModelPath(urdf_path, self.verbose)
-        self.urdf_path = join(self.model_path, urdf_path)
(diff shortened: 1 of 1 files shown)


cmastalli (commented):
Should we always give priority to build with the URDF model? 

I suspect that might be cases where we want to build through the SDF model. If so, the loader needs to reflect the wish of giving specific priorities.

A part of this comment, the rest looks fine for me.

nim65s:
I was assuming that we will never have a model with both an URDF model and a SDF model

python/example_robot_data/robots_loader.py
```diff
@@ -50,19 +50,29 @@ class RobotLoader(object):
...
 
     def __init__(self):
-        urdf_path = join(self.path, self.urdf_subpath, self.urdf_filename)
-        self.model_path = getModelPath(urdf_path, self.verbose)
-        self.urdf_path = join(self.model_path, urdf_path)
-        self.robot = RobotWrapper.BuildFromURDF(self.urdf_path, [join(self.model_path, '../..')],
-                                                pin.JointModelFreeFlyer() if self.free_flyer else None)
+        if self.urdf_filename:
+            df_path = join(self.path, self.urdf_subpath, self.urdf_filename)
+            builder = RobotWrapper.BuildFromURDF
```
cmastalli, line 65: I would add a warning message if the developer has defined `sdf_filename`. I would include the following code:
```python
  if self.sdf_filename:
    print("Warning: we are building the robot model from the URDF")
    print("If you wish to build the model from the SDF", then do not define the 'urdf_filename'")
```

proyan:
ping :)
Could we please finish this PR, so that #100 could also be handled?

python/example_robot_data/robots_loader.py
```diff
@@ -50,19 +50,31 @@ class RobotLoader(object):
...
+            if self.sdf_filename:
+                raise AttributeError("Please choose between URDF *or* SDF")
+            df_path = join(self.path, self.urdf_subpath, self.urdf_filename)
+            builder = RobotWrapper.BuildFromURDF
+        else:
+            df_path = join(self.path, self.sdf_subpath, self.sdf_filename)
+            try:
+                builder = RobotWrapper.BuildFromSDF
+            except AttributeError:
+                raise ImportError("Building SDF models require pinocchio >= 2.9.2")
```
proyan, line 73:
```suggestion
                raise ImportError("Building SDF models require pinocchio >= 2.9.1")
```

jcarpent: I would suggest using Pinocchio >= 3.0.0. 
This is much more simple, as 2.9.x won't be released on Conda, APT, etc.

nim65s: Ok for 3. It's a little lie, but it won't hurt users who have 2.9 :)

PR #108, opened 2021-11-04
fix inertia on finger_edu & a1, ref #68

nim65s:


jcarpent (commented):
One of the main issues is that Forward and Inverse dynamics algo will not work.

cmastalli:
It would be useful to include a tiny inertia (e.g., 1e-6) as Pybullet simulator doesn't like body with no inertia.

nim65s:
@jcarpent : Those links are only here for the collisions, so those values should not mess with dynamics algo.

@cmastalli : what do you mean by "doesn't like" ? This modification is here to prevent pybullet to assign a 1kg default mass to those links which have none. Did #68 raise any issue for you ?

cmastalli (approved):

PR #109, opened 2021-11-10
allow customization of model path

nim65s:
Hi,

In cases where a user wants to use a custom URDF while keeping other parts of the robot (srdf and/or meshes), we should be able to provide directly the path to that custom URDF.

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -58,6 +58,7 @@ class RobotLoader(object):
     has_rotor_parameters = False
     free_flyer = False
     verbose = False
+    model_path = None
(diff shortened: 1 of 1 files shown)


python/example_robot_data/robots_loader.py
```diff
@@ -58,6 +58,7 @@ class RobotLoader(object):
     has_rotor_parameters = False
     free_flyer = False
     verbose = False
+    model_path = None
```
cmastalli, line 61: Perhaps, it is a good idea to start adding some documentation for these hyper-parameters.

cmastalli (approved):

PR #111, opened 2021-11-25
Add continuous (SO2) variant of double_pendulum

ManifoldFR:
This PR adds a second version of the double pendulum URDF where the joints are defined as `continuous` (SO2) instead of revolute, and the appropriate options in the `load()` function.


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you
can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

ManifoldFR:
Edit: rebased on Gepetto:devel branch

python/example_robot_data/robots_loader.py
```diff
@@ -483,8 +483,16 @@ class DoublePendulumLoader(RobotLoader):
     urdf_subpath = "urdf"
 
 
-def loadDoublePendulum():
+class DoublePendulumContinuousLoader(RobotLoader):
```
nim65s, line 486: This class should inherit from `DoublePendulumLoader` and only override its `urdf_filename`

python/example_robot_data/robots_loader.py
```diff
@@ -483,8 +483,16 @@ class DoublePendulumLoader(RobotLoader):
...
+    path = "double_pendulum_description"
+    urdf_filename = "double_pendulum_continuous.urdf"
+    urdf_subpath = "urdf"
+
+
+def loadDoublePendulum(continuous=False):
```
nim65s, line 492: This load function is deprecated, will be removed, and shouldn't be used. Therefore, there is no need to update it

nim65s (changes requested):
This looks useful, thanks for the PR !

Could you take care of those 2 comments, and add a unit test next to https://github.com/Gepetto/example-robot-data/blob/master/unittest/test_load.py#L53 ?

python/example_robot_data/robots_loader.py
ManifoldFR: Will fix asap

python/example_robot_data/robots_loader.py
ManifoldFR: Alright!

ManifoldFR:
> This looks useful, thanks for the PR !
> 
> Could you take care of those 2 comments, and add a unit test next to https://github.com/Gepetto/example-robot-data/blob/master/unittest/test_load.py#L53 ?

I took care of the comments and added a unit test (and ran the `test_load.py` file again)

cmastalli (commented):
As far I understand, this PR proposes to include a slightly modified URDF for the double pendulum. The only modification that I see is that the joints are continuous rather than revolute.

My question is do we really need to create an extra URDF?
An alternative solution will be to modify the Pinocchio model.

nim65s (approved):
Looks good to me, thanks

nim65s:
> My question is do we really need to create an extra URDF? An alternative solution will be to modify the Pinocchio model.

Both solutions are fine for me. I mean, I agree it's not super nice to duplicate the URDF, but, on the other hand, changes to the source will be trivial to apply to the copy.

At the same time, it's not super nice to tie robot models to pinocchio, especially when we want to benchmark pinocchio vs something else on a particular model, but, on the other hand, well… We are already clearly privileging pinocchio for now, in the python utils and in the unit tests.

nim65s:
I think it is relevant to have both constraint and unconstraint double pendulums, especially for teaching and tutorials, so this would really fit with the purpose of this repo :)
This is probably not true for most other models we have.

jcarpent (approved):

jcarpent (commented):
A small comment, it would have been possible to not duplicate the URDF file but rather to change the type of joints after loading ;)

cmastalli (approved):
I think we could go ahead with this request.

ManifoldFR:
Hi, can we merge this PR? Since it was already approved

nim65s:
Sure, sorry for the delay

ManifoldFR:
Thanks !

PR #115, opened 2021-12-08
added Articulated soft robot loader and description

spykspeigel:
I have added a 2DOF articulated soft robot with its description and loader.


jcarpent (changes requested):
Thanks @spykspeigel for sharing this model.
Before merging this PR, it is important for us to understand:
- where does this model come from? Can it be provided by anybody?
- does this model will be useful to the whole community of Pinocchio, Crocoddyl and other software users?

By the way, you should only provide the minimal set of files to make things loadable by the parsers. At the time being, there are a lot of files that are not required (CMake files for instance).

cmastalli:
> Thanks @spykspeigel for sharing this model.
> Before merging this PR, it is important for us to understand:
> - where does this model come from? Can it be provided by anybody?
> - does this model will be useful to the whole community of Pinocchio, Crocoddyl and other software users?
> 
> By the way, you should only provide the minimal set of files to make things loadable by the parsers. At the time being, there are a lot of files that are not required (CMake files for instance).

@spykspeigel is working with me. This robot will eventually be useful for Crocoddyl, when we finished our ongoing work.

The robot is originally defined by Pisa. I guess we are authorized to publish this data, right @spykspeigel ?

Furthermore, as pointed out by @jcarpent, we really need the smallest set of files, which are the urdf and dae files.

You need to remove the xacro, launch, STL files.

python/example_robot_data/__init__.py
```diff
@@ -1,4 +1,4 @@
 # flake8: noqa
 from .robots_loader import (ROBOTS, getModelPath, load, load_full, loadANYmal, loadDoublePendulum, loadHector, loadHyQ,
                             loadICub, loadIris, loadKinova, loadPanda, loadRomeo, loadSolo, loadTalos, loadTalosArm,
-                            loadTalosLegs, loadTiago, loadTiagoNoHand, loadUR, readParamsFromSrdf)
+                            loadTalosLegs, loadTiago, loadTiagoNoHand, loadUR, readParamsFromSrdf, loadAsrTwoDof)
```
nim65s, line 4: You can remove this

python/example_robot_data/robots_loader.py
```diff
@@ -135,6 +135,14 @@ class TalosLoader(RobotLoader):
     srdf_filename = "talos.srdf"
     free_flyer = True
 
+class AsrTwoDofLoader(RobotLoader):
+    path = 'asr_twodof_description'
+    urdf_filename = "TwoDofs.urdf"
+    urdf_subpath = "urdf"
+
+def loadAsrTwoDof():
```
nim65s, line 143: There is no need to create a deprecated function. The Loader class is enough

nim65s (changes requested):
In addition to other comments, and the details about the load function, could you add a README documenting the robot provenance and its origin license ?

spykspeigel:
Dear All, Thanks for the comments. I have made the suggested changes i.e. removed the unnecessary files, added readme and license files. Kindly let me know if it needs any further changes.

nim65s (changes requested):
Thanks for the fixes, it's cleaner now !

I still see 2 issues:
1. the CI is failing. This is because you added the robot under the name `asr_twodof`, and tried to uni-test the robot `asr_two_dof`. Please pick one :)
2. you are also adding `simplecube_description`. If this not intended, you should remove this folder. Otherwise, you should add a loader and a unit test for this robot too. And maybe update the title of this pull request

robots/asr_twodof_description/urdf/TwoDofs.urdf
```diff
@@ -0,0 +1,142 @@
...
+    </gazebo>
+    <gazebo>
+        <plugin name="advanced_vsa_j2" filename="libadvanced_plugin.so">
+            <joint>J2</joint>
+            <control_type>2</control_type>
+            <pub_eltau>false</pub_eltau>
+            <pub_state>false</pub_state>
+            <sub_ext_tau>false</sub_ext_tau>
+        </plugin>
+    </gazebo>
```
cmastalli, line 140: I think we should remove these plugins as they are not standard ones.
@nim65s  and @jcarpent - do you agree or not on this topic?

robots/simplecube_description/urdf/simplecube.urdf
```diff
@@ -0,0 +1,92 @@
...
+
+    <gazebo>
+        <plugin name="advanced_vsa_j1" filename="libadvanced_plugin.so">
+            <joint>J1</joint>
+            <control_type>1</control_type>
+            <pub_eltau>false</pub_eltau>
+            <pub_state>true</pub_state>
+            <sub_ext_tau>false</sub_ext_tau>
+        </plugin>
+    </gazebo>
```
cmastalli, line 91: same here

cmastalli (changes requested):
Please also handle the issue about the `simplecube` raised by @nim65s in https://github.com/Gepetto/example-robot-data/pull/115#pullrequestreview-829517409

spykspeigel:
@nim65s the urdf file in `two_dof_description` depends on the dae files from `simplecube_description`. So I have moved the corresponding files in folder `qb_meshes` to the `two_dof_description` and removed `simplecube_description` .

cmastalli (approved):

nim65s (approved):

PR #117, opened 2022-02-18
Release/3.13.1

nim65s:


nim65s:
I mean these are not real links on the robot, as they don't have a body, with shape and inertia, right ?

cmastalli:
> I mean these are not real links on the robot, as they don't have a body, with shape and inertia, right ? edit: `ground` as a mesh. But what about `Tip` ?

That's correct. I think `ground` can be removed and `tip` might be useful for being able to define the robot's tip.
@spykspeigel -- do you agree? could you handle this issue asap?

spykspeigel:
ground link has a mesh and Tip link can be useful.
So as I understand I shouldn't modify anything?

nim65s:
I don't see how the `Tip` link could be useful. It doesn't have anything more than the `EE` link, has it ?
On the other hand, I can see how this `Tip` link can break simulations: it doesn't define a mass, so the pybullet simulator will by default assume it has a mass of 1.0 kg, which is plain wrong.

spykspeigel:
@nim65s  Thanks a lot for handling this.

PR #118, opened 2022-03-29
Add simplified double pendulum

justinberi:
Adds a standalone double pendulum urdf . https://github.com/loco-3d/crocoddyl/pull/1058

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -16,17 +16,23 @@ def _depr_msg(deprecated, key):
 def getModelPath(subpath, printmsg=False):
     source = dirname(dirname(dirname(__file__)))  # top level source directory
     paths = [
-        join(dirname(dirname(dirname(source))), 'robots'),  # function called from "make release" in build/ dir
-        join(dirname(source), 'robots'),  # function called from a build/ dir inside top level source
+        join(dirname(dirname(dirname(source))),
+             'robots'),  # function called from "make release" in build/ dir
+        join(dirname(source), 'robots'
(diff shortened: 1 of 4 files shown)


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master. If this wasn't intentionnal, you
can change the base branch of this pull request to devel (No need to close it for that). Best, a bot.

nim65s:
Thanks for this simple model. 

Just to be explicit, about this, is this file an original work from you ? And if this the case, is BSD-2-Clause license OK for you ? 

Otherwise, where does this come from and what is its original license ?

Also, it would be nice to add a loader after this one: https://github.com/Gepetto/example-robot-data/blob/2b998b16a2544cd2392ee6f9cc075362e6259974/python/example_robot_data/robots_loader.py#L494, add this loader to the `ROBOTS` dict, and add a unit test after this one: https://github.com/Gepetto/example-robot-data/blob/2b998b16a2544cd2392ee6f9cc075362e6259974/unittest/test_load.py#L57

Can you do this, or would you prefer me to handle that ?

justinberi:
No problems. Happy to contribute!

> Just to be explicit, about this, is this file an original work from you ? And if this the case, is BSD-2-Clause license OK for you ?

It is modified from `double_pendulum.urdf`. To remove the meshes (and also added a frame for the endeffector). Yes happy with BSD-2-Clase license. I am happy to inherit the license. 

Because it is so heavily modified, would you like it in a different directory?

Added loader and unittests.

nim65s:
Thanks a lot !

Issue #112, opened 2021-11-30
Synchronizing simple_humanoid in pinocchio/models and example_robot_data


proyan:
I suggest that we move pinocchio's `simple_humanoid` from `pinocchio/models` to `example_robot_data`. It would make the sample humanoid available for other packages as well. Any views?

jcarpent:
Totally fine on my side.

proyan:
Ah my bad! It looks like it is already there, but is duplicated in pinocchio! This issue is ill-placed, I'll close it.

proyan:
I'll create a PR synchronizing the models from pinocchio and here, let's change the direction of this issue :)

cmastalli:
It sounds good for me as well.
Thanks @proyan for taking care of it :)

cmastalli:
Is this issue sorted out?

nim65s:
Actually, I added this robot in #69,  8 months before this PR, so I'm not sure what this PR is about.

proyan:
#120 handles the changes I was proposing

PR #120, opened 2022-04-01
simple_humanoid: add mesh, update srdf with rotor info, urdf with coliison info

proyan:
Deals with #112

### DIFF
--- a/robots/simple_humanoid_description/urdf/simple_humanoid.urdf
+++ b/robots/simple_humanoid_description/urdf/simple_humanoid.urdf
@@ -21,7 +21,7 @@
     </collision>
     <collision name="box">
       <geometry>
-        <mesh filename="package://simple_humanoid_description/box.stl" />
+        <mesh filename="package://example-robot-data/robots/simple_humanoid_description/meshes/box.stl" />
       </geometry>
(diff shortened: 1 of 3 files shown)


robots/simple_humanoid_description/urdf/simple_humanoid.urdf
```diff
@@ -14,6 +14,22 @@
       <mass value="27"/>
       <inertia ixx="1" ixy="0" ixz="0" iyy="1" iyz="0" izz="1" />
     </inertial>
+    <collision name="test">
+      <geometry>
+        <cylinder radius="1" length="1"/>
+      </geometry>
+    </collision>
```
cmastalli, line 21: Is this a mistake?
Btw, I am not sure if we can add two collision objects in the URDF.

proyan: We can have multiple collision objects, and multiple visual objects.

cmastalli: Ok, thanks for the clarification.

cmastalli (approved):

PR #121, opened 2022-04-01
cassie/cassie_v2 sdf: remove first primatic joint, add color information

proyan:


python/example_robot_data/robots_loader.py
```diff
@@ -161,11 +175,27 @@ def loadANYmal(withArm=None):
 
 class CassieLoader(RobotLoader):
     path = 'cassie_description'
-    sdf_filename = "cassie_v2.sdf"
+    if tuple(int(i) for i in pin.__version__.split('.')) > (2, 9, 1):
+        sdf_filename = "cassie.sdf"
+    else:
+        sdf_filename = "cassie_v2.sdf"
```
nim65s, line 181:
```suggestion
    sdf_filename = "cassie.sdf" if tuple(int(i) for i in pin.__version__.split('.')) > (2, 9, 1) else "cassie_v2.sdf"
```

nim65s:
Don't worry about the CI.

If loading cassie is working for you, I think we can merge and continue with the release

python/example_robot_data/robots_loader.py
nim65s: (I don't think you can actually execute code in the class declaration, but you can execute code in the parameter declaration… But I might be wrong)

nim65s: my bad, this works, I wasn't aware of it

proyan:
I'm finished with the changes, you could merge it now.
Thanks for your help

PR #134, opened 2022-08-07
Fix anymal robot preview in MeshCat (Tree.js)

jcarpent:


cmastalli:
I checked that this works in Meshcat; however, I am unable to do so in GV and RVIZ.
Have you checked how the robot displays in GV and RVIZ?
If so, can you share pictures?

nim65s:
before
after

nim65s:
I rebased to avoid the conflict

jcarpent:
So it works @nim65s ?

cmastalli:
> in gepetto-viewer, before:
> before
> after:
> after

It has lower quality, could be possible to add back the missed colours?

cmastalli:
> what ? the real robot really has those violet stuff ? It looks like an artifact that @jcarpent version is fixing for me. Here is another "before" view:
> front

Sorry I am not sure what you meant. Instead I meant, the colours in the motors and fan are gray. Furthermore, there are plates or covers that are blue. Additionally, there is a texture in the body.

However, answering your questions. Yes, the ANYmal robot looks pretty much that the original picture. These texture were added by ANYbotics.

jcarpent:
Honestly, I don't know how to recover from this issue.
The main purpose on my side is to have robots which can be displayed in the viewers associated to Pinocchio.

The current anymal robot cannot be displayed in Meshcat. The fix that we suggest here fixes this issue, yet with some minor degradation.
As gepetto-viewer is no more deeply maintained, I would suggest using MeshCat as the main interface, as it is also multiplatform, unlike GV.

Could we then merge this PR and open a pending issue concerning the fix for retrieving the blue colors?

cmastalli:
> Honestly, I don't know how to recover from this issue. The main purpose on my side is to have robots which can be displayed in the viewers associated to Pinocchio.
> 
> The current anymal robot cannot be displayed in Meshcat. The fix that we suggest here fixes this issue, yet with some minor degradation. As gepetto-viewer is no more deeply maintained, I would suggest using MeshCat as the main interface, as it is also multiplatform, unlike GV.
> 
> Could we then merge this PR and open a pending issue concerning the fix for retrieving the blue colors?

We use this model in Rviz as well, and for many of us, we wish to still have good quality videos, e.g.,
https://youtu.be/mlrTdiOA1hM

Generally, I don't believe it is a good development strategy to degrade others, for the sake of having extra features. Additionally, it is not clear (at least for me) why we need to do this here, instead of solving the issue in `Tree.js`. Clarity of this aspect might help us to evaluate the proposition of a PR.

Avoiding long-to-release solutions (that could be the case in extending `Tree.js`) is always desirable. If you tell me that this this is case, then I suggest an alternative solution that doesn't compromise people interested in using other visualisation tools. The alternative solution is to have two DAE files while keeping the current one as default. Then, third party software (e.g., Pinocchio) can develop internally (in Pinocchio) a logic that choose the desired DAE.

All these are patches over patches, but it could be faster to release.

I understand your frustration in having this merge, however, I am trying to play the role of devil's advocate :)

jcarpent:
> Additionally, it is not clear (at least for me) why we need to do this here, instead of solving the issue in `Tree.js`

Do you know how to do that? Are you willing to do that?

cmastalli:
> > Additionally, it is not clear (at least for me) why we need to do this here, instead of solving the issue in `Tree.js`
> 
> Do you know how to do that? Are you willing to do that?

No, I don't.

nim65s:
If none of us has the time to get a version of this mesh that works well enough everwhere, we can duplicate it, and let the user choose.

But I'll try to get a more clear look at this Tomorrow, and open issues on Tree.js and/or the original anymal repo if needed.

wxmerkt:
Can we use obj for MeshCat?

jcarpent:
yes

wxmerkt:
I am looking into a fix for this. I have a workaround that detects and "fixes" broken DAEs on the fly rather than omitting them from Meshcat/THREE.js. It requires some more testing before upstreaming

wxmerkt:
I've got a workaround that allows loading the current mesh in the repository in Meshcat, see upstream PR: https://github.com/rdeits/meshcat/pull/139

Using this, we get the following:

image

Let's see if the upstream PR gets merged.

And with some extra changes to meshcat-python that I am working on we could get texture support:

image

Alternatively, I have a new mesh that bakes in the colours and displays correctly but it's significantly larger so I won't commit it.

cmastalli:
I'll close this PR. Feel free to re-open if required, @jcarpent!

PR #145, opened 2022-12-08
Fix Kinova mesh materials and remove Anymal mounting block

wxmerkt:
I am working on textures (and fixing #134) and noticed that the Kinova URDF had a few meshes that did not contain materials. It also loaded the Anymal B mounting block which shouldnt be in the robot standalone URDF.

With these changes, and with the yet unfinished Meshcat changes [needs some more debugging to fix #134], we can visualise the robot as:

image


nim65s (approved):

cmastalli (approved):

cmastalli:
Thanks @wxmerkt!

I will proceed with merging this PR.

PR #147, opened 2022-12-12
Fix Panda inertials, add SRDF

### LINKED ISSUE #122: Dynamic properties of Panda URDF are missing
Equally, the `q0` is all zeros, which is a singular configuration. This is the root cause for the issue described by @spykspeigel in https://github.com/loco-3d/crocoddyl/issues/1064

Cf 
image

wxmerkt:
Fixes #122


robots/panda_description/urdf/panda.urdf
```diff
@@ -82,46 +84,48 @@
...
     <link name="panda_link4">
-        <inertial>
-          <mass value="0"/>
-          <origin rpy="0 0 0" xyz="0 0 0"/>
-          <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0"/>
-        </inertial>
         <visual>
             <geometry>
-                <mesh filename="package://example-robot-data/robots/panda_description/meshes/visual/link4.dae" />
+                <mesh filename="package://franka_description/meshes/visual/link4.dae"/>
```
cmastalli, line 105: We should use the path relative to example-robot-data.

There are other places where we need to revert back this.

robots/panda_description/urdf/panda.urdf
```diff
@@ -225,56 +231,121 @@
...
         <visual>
             <geometry>
                 <mesh filename="package://example-robot-data/robots/panda_description/meshes/visual/finger.dae" />
             </geometry>
         </visual>
+        <!-- screw mount -->
         <collision>
+            <origin rpy="0 0 0" xyz="0 18.5e-3 11e-3"/>
             <geometry>
-                <mesh filename="package://example-robot-data/robots/panda_description/meshes/collision/finger.stl" />
```
cmastalli, line 242: Is this intended to be? If so, why?

robots/panda_description/urdf/panda.urdf
```diff
@@ -225,56 +231,121 @@
...
+            <geometry>
+                <box size="22e-3 8.8e-3 3.8e-3"/>
+            </geometry>
+        </collision>
+        <!-- diagonal finger -->
         <collision>
-            <origin rpy="0 0 3.14159265359" xyz="0 0 0" />
+            <origin rpy="-0.5235987755982988 0 0" xyz="0 -15.9e-3 28.35e-3"/>
             <geometry>
-                <mesh filename="package://example-robot-data/robots/panda_description/meshes/collision/finger.stl" />
```
cmastalli, line 261: Same here

cmastalli (commented):
Thank @wxmerkt for taking care of this issue.
I have left some comments about the PR.

robots/panda_description/urdf/panda.urdf
wxmerkt: The upstream Gazebo simulation URDF now uses a box. I believe the finger was a mesh that was approximately a box previously, so no it's cheaper (and numerically more stable for simulation) to use a primitive shape

wxmerkt:
> Also, I think this is a good thing to update this inertia, but I'm afraid this might break benchmarks or other reproducible tests relying on the current values. But I guess we'll only have to advertise this correctly in release notes.

I don't think we need to worry about this: We haven't used it in benchmarks yet to my knowledge - and _couldn't_ due to the zero inertials - i.e. that was the reason behind #122 and the original issue upstream in Crocoddyl about failed optimisation (due to zero inertials).

cmastalli (approved):
I don't have any further concern.

nim65s (approved):
You're right, we couldn't, it's perfect for me then :)

PR #149, opened 2022-12-14
[panda_description] Update meshes from franka_ros to get textures

### LINKED ISSUE #146: Panda mesh textures broken
The meshes added in 083e77ecb62c34c5df71e7231d0c59fe6223c330 (fixed with MeshLab) do not display nicely (this is with RViz):

image

We should investigate how to restore textures while still being able to visualise in Meshcat

wxmerkt:
Resolves #146

No changes to rendering with the currently published meshcat version. But once my current DAE handling fixes are ready to be upstreamed it will look like this:

image


cmastalli (commented):
Do you know what is the specific change from Blender that fixes this?

cmastalli (approved):

PR #140, opened 2022-11-09
Adding allegro hands (left and right)

lmontaut:
Python api: `example-robot-data.load("allegro_right_hand" or "allegro_left_hand")` .

PR #155, opened 2023-02-09
Adding Allegro with free base

lmontaut:
Hi, I forgot to add the allegro hand with a free-flyer at its base when doing the PR #140.


lmontaut:
As discussed with @jcarpent adding this is redondant. Either we make the Allegro hand free by default (new PR) or in Pinocchio we can use `appendModel` to fuse 2 models (allegro fixed + a free-flyer).

PR #158, opened 2023-02-13
go1: fix inertia of fake camera links for pybullet

nim65s:
while here, little follow up on #156


cmastalli (approved):
We have recently included B1. Do you know if this action is needed there as well?

nim65s:
If you install pybullet and unit tests are passing, no more actions are needed, and it looks like this is the case for me.

Issue #164, opened 2023-04-23
File too large for go1 robot


jcarpent:
Looking at the mesh files of the GO1 robot (https://github.com/Gepetto/example-robot-data/tree/devel/robots/go1_description/meshes), at least `trunk.dae` is too large:
```bash
 2.8M Feb 10 14:55 calf.dae
  69K Feb 10 14:55 depthCamera.dae
 2.9M Feb 10 14:55 hip.dae
 1.2M Feb 10 14:55 thigh.dae
 1.5M Feb 10 14:55 thigh_mirror.dae
  88M Feb 10 14:55 trunk.dae
 279K Feb 10 14:55 ultraSound.dae
 ```
 We should think about reducing the size before any release.

cmastalli:
@Sergim96 -- could you address this issue?

jcarpent:
Seems to be solved in release 4.0.6

PR #166, opened 2023-05-03
Release/4.0.6

nim65s:


wxmerkt:
@nim65s We probably want to address #164 prior to a release

PR #169, opened 2023-05-12
always try to import MODEL_DIR and DATA_SOURCE_DIR

fabinsch:
Hello, I'm currently using `example-robot-data` to have access to different urdf files. But in my case, I do not want to use the `RobotLoader`, but use `pinocchio` to build my model from URDF directly. I noticed, that you can do
```
import example_robot_data as erd
erd.getModelPath()  # of course this fails
erd.path.EXAMPLE_ROBOT_DATA_MODEL_DI
```
but you cannot directly access `erd.path` after loading the python module. I think it can be useful in general to have access to both `EXAMPLE_ROBOT_DATA_MODEL_DIR` and `EXAMPLE_ROBOT_DATA_SOURCE_DIR` even without calling a specific function before.

### DIFF
--- a/python/example_robot_data/robots_loader.py
+++ b/python/example_robot_data/robots_loader.py
@@ -5,6 +5,10 @@ import numpy as np
 import pinocchio as pin
 from pinocchio.robot_wrapper import RobotWrapper
 
+try:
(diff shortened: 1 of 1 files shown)


nim65s:
I'm not sure I understand what you need, but anyway this is not going to hurt anybody else :)

PR #180, opened 2023-07-30
Supported pinocchio with cppadcg installation

cmastalli:
This enables us to install example-robot-data when Pinocchio has been installed with CppADCogen


hrp2-14:
Hi ! This project doesn't usually accept pull requests on the main branch.
If this wasn't intentionnal, you can change the base branch of this PR to devel
(No need to close it for that). Best, a bot.

cmastalli (commented):
[cleanup] Fixed cmake format

nim65s:
This should be fixed in pinocchio packaging, not here, and in each package dependent on pinocchio

cmastalli:
> This should be fixed in pinocchio packaging, not here, and in each package dependent on pinocchio

Agreed! I created an issue and compromise to handle it when is possible. Is this enough for now? Or would you prefer to revert this PR?

PR #193, opened 2023-11-29
add changelog

nim65s:


cmastalli (approved):
Thanks, @nim65s !

I am working now to adopt this in Crocoddyl. Saying this to avoid duplication of effort (cc @jcarpent)

PR #201, opened 2024-01-19
Added the Unitree's Z1 robot, improved colors in Unitree robots, and defined the B1_Z1 robot.

cmastalli:


nim65s:
b1 and go1 are 37Mo and 36Mo, the heaviest of the repo. I really think we should optimize before merging. But now this is merged, so we'll have to rewrite history to avoid having this burden in the git history forever.

cmastalli:
Is this was introduced by my changes? I just added colors in blender.

How much was it before?

cmastalli:
After checking the sizes of mesh files for b1 and go1 robots before my changes, I can see the same file sizes. Moreover, these files were added in December 2022, which prohibits us from rewriting history.

It is worth reducing the file sizes. Do you have any advice for doing so? (@nim65s)

PR #202, opened 2024-01-19
Added quadrotors with 2 and 5 dofs arm

cmastalli:
This quadrotor models were developed by @PepMS (Thanks!)

@PepMS -- your repo doesn't have details on license. Please let me know if I need to define the license of these URDFs.


PepMS:
Hi, sorry for arriving late on this.

I never thought these designs would go that far; that is why I didn't add a license to them. Besides, my fork of example-robot-data repo turned into a group repository for descriptions of our robots, more refined and adapted to our needs.

For the specific issue of the license, let me discuss it with the group, and I will come back to you.

One last important thing to note is that these designs do not contain rotors as links. Therefore, the multirotor is just a rigid body. If you pretend to use this, e.g., in Crocoddyl you need to know how the different rotor forces are located within the multicopter base. This info is something that the URDF does not contain and that I kept in a separate repo. To keep things simple, I would suggest adding this info as a comment somewhere in the URDF files (unless you want to specify the rotors as proper URDF links).

cmastalli:
> Hi, sorry for arriving late on this.
> 
> I never thought these designs would go that far; that is why I didn't add a license to them. Besides, my fork of example-robot-data repo turned into a group repository for descriptions of our robots, more refined and adapted to our needs.
> 
> For the specific issue of the license, let me discuss it with the group, and I will come back to you.
> 
> One last important thing to note is that these designs do not contain rotors as links. Therefore, the multirotor is just a rigid body. If you pretend to use this, e.g., in Crocoddyl you need to know how the different rotor forces are located within the multicopter base. This info is something that the URDF does not contain and that I kept in a separate repo. To keep things simple, I would suggest adding this info as a comment somewhere in the URDF files (unless you want to specify the rotors as proper URDF links).
> 
> @cmastalli

Indeed, I was thinking of defining links for propellers. This would be useful in Crocoddyl as well. In future, I am planning to extend our API by defining a thruster in a particular frame.

It would be fantastic if you can help me with this task. Can you create a PR with the propeller links and license?

Including the propeller coefficients inside the SDRF can be a potential future effort. This would allow us to create functions (e.g., inside Pinocchio) to parse this information. @nim65s and @jcarpent -- I would love to hear your feedback on this point as well.

PR #209, opened 2024-03-28
Add model of PR2 robot

florent-lamiraux:


nim65s:
This is missing loader and unittests, but more importantly we need to squash those commits to avoid adding a ton of unused meshes to the git history

nim65s:
Also, with a bit of optipng, we can reduce this from 8.6M to 7.8M

nim65s:
Arg, wrong git command, sorry @florent-lamiraux

PR #225, opened 2024-07-16
Enable jrl-cmakemodules workspace compatiblity

jorisv:


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "cmake"]
 	path = cmake
-	url = https://github.com/jrl-umi3218/jrl-cmakemodules.git
+	url = https://github.com/jorisv/jrl-cmakemodules.git
```
cmastalli, line 3: Any reason to do this?

CMakeLists.txt
```diff
@@ -64,23 +69,27 @@ compute_project_args(PROJECT_ARGS LANGUAGES CXX)
 project(${PROJECT_NAME} ${PROJECT_ARGS})
 
 if(BUILD_PYTHON_INTERFACE)
-  add_project_dependency(eigenpy 3.0.0 REQUIRED)
-  add_project_dependency(pinocchio 2.7.0 REQUIRED)
```
cmastalli, line 68: Pinocchio is a project dependency. Any reason to do this?

unittest/CMakeLists.txt
```diff
@@ -1,13 +1,6 @@
-if(NOT pinocchio_FOUND)
-  message(
-    FATAL_ERROR
-      "Pinocchio is required to build unit tests\n"
-      "please either install Pinocchio or reconfigure without unit tests (-DBUILD_TESTING=OFF)"
-  )
-endif(NOT pinocchio_FOUND)
```
cmastalli, line 7: Same comment here.

.gitmodules
jorisv: It's linked to this pull request : https://github.com/jrl-umi3218/jrl-cmakemodules/pull/701
submodule will be changed when jrl-cmakemodules PR will be merged.

unittest/CMakeLists.txt
jorisv: The only test is a Python test. Then, checking if find_package(pinocchio) have succeed to detect if Pinocchio Python binding will not help.

More generally, I don't think that CMake should check for Python dependencies with CMake modules. Python dependencies should be managed by a package manager (pip, conda, ...).

CMakeLists.txt
cmastalli: I understand. Imposing this dependency ensures that users have installed Pinocchio from the user side. This is the only requirement where (I think) we should consider here.

jorisv:
@cmastalli Did the modifications are fine to you ?

cmastalli:
> @cmastalli Did the modifications are fine to you ?

I think we should not remove the in https://github.com/Gepetto/example-robot-data/pull/225#discussion_r1679691018. However, I would like to hear from @nim65s on this topic.

jorisv:
@cmastalli I revert the change about add_project_dependency({eigenpy,pinocchio}).

If it's fine to you, you can merge.

unittest/CMakeLists.txt
cmastalli: Hi @jorisv!

This is the last comments. Please ping us when it is done! Thanks!

jcarpent:
@cmastalli ping to merge this PR if you are fine.

cmastalli:
> @cmastalli ping to merge this PR if you are fine.

Hi @jcarpent -- We're just missing to address this comment: https://github.com/Gepetto/example-robot-data/pull/225#discussion_r1692859261.

This should take no time to do it.

unittest/CMakeLists.txt
cmastalli: Could you provide details on why those cases fail? In any case, I'll let @nim65s decide this. He is more competent in judging your suggestion. Thanks!

nim65s: We only have python unittest, and when python is active, there is a `add_project_dependency(pinocchio 2.7.0 REQUIRED)`, so the only case where things could fail would be if someone installs pinocchio without python interface and try to run example-robot-data python unittest.
This case can be dismissed, so I think we can remove this error message.

nim65s (approved):

PR #233, opened 2024-09-05
Removed warning obtained with Pinocchio 3

cmastalli:


nim65s (approved):

python/example_robot_data/robots_loader.py
```diff
@@ -333,7 +333,10 @@ def __init__(self, verbose=False):
 
         # q2 = self.robot.q0[:19]
         for f in m1.frames:
-            if f.parent < legMaxId:
+            if tuple(map(int, pin.__version__.split("."))) >= (3, 0, 0):
```
nim65s, line 336: list comprehensions are slightly more idiomatic than map in python, so it would also work with `tuple(int(i) for i in pinocchio.__version__.split('.'))

cmastalli: Done

PR #239, opened 2024-10-09
Add go2 description

edantec:


nim65s (changes requested):
This is adding 50M to the repo, which is already way too big.

No other robot folder is that big. 
PR2 is 1.7M, talos 7.4M for comparison.

Can you try to reduce the meshes ? Surely we don't need that level of detail for examples and unit tests

edantec:
Meshes now weight 5.5 Mo, is it light enough? I used Blender to decimate the geometries.

nim65s:
Way better, thanks a lot :D
Can you just double check they appear correctly on our viewers: meshcat / gepetto-gui / rviz / panda.
Maybe not all of those, but at least the 2 your prefer

After that, you also need to add an unit test, and this is good for me (we must squash the commits, though, otherwise we'll still have those 50M added in the repo)

cmastalli (commented):
I have two comments somehow related:
  1. Can we reuse the mesh files from Go1? Are they fundamentally different?
  2. Can we reuse the SRDF file from Go1?

It seems that point 2 is quite doable. Name of links and bodies look similar.

edantec:
Capture d’écran du 2024-10-10 14-41-44
Capture d’écran du 2024-10-10 14-41-28

Here is what the go2 looks like with simplified meshes. What do you think? It seems good enough to me.

edantec:
> I have two comments somehow related:
> 
>     1. Can we reuse the mesh files from Go1? Are they fundamentally different?
> 
>     2. Can we reuse the SRDF file from Go1?
> 
> 
> It seems that point 2 is quite doable. Name of links and bodies look similar.

I actually reused the srdf file from go1 to create the one for go2. Standing position is different though, I modified the SRDF accordingly.
As for the mesh files, they are quite different.

nim65s:
Yes, I think this is good enough, thanks !
Can you just add a unit test for the loader with the right nq / nv ?

edantec:
> Yes, I think this is good enough, thanks ! Can you just add a unit test for the loader with the right nq / nv ?

Done. I noticed however that the test is not passing for the pr2 robot.

nim65s:
pr2 test seems ok on CI tho

PR #242, opened 2024-10-29
Ahaffemayer/add panda collision

ArthurH91:
Hello, here's a PR to have the panda robot with the capsules in its collision model directly in example robot data. 

In this Pr is included:

- A new URDF with the cylinders & spheres description of the Panda Robot.
- A subclass of RobotLoader which create a robot model from the URDF, then transform all the spheres/cylinders couple into capsules. 

I decided to have the hand and the fingers described as boxes here because it's more faithful to the real representation. However, on your input, I can change and transform it into capsules.

Here's a superposed view of the visual model (with meshes) and the collision model: 

Screenshot from 2024-10-29 13-06-55


hrp2-14:
Hi ! This project doesn't usually accept pull requests on the main branch.
If this wasn't intentionnal, you can change the base branch of this PR to devel
(No need to close it for that). Best, a bot.

ArthurH91:
Here's the modified version with hand and fingers as capsules. 

image

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +459,78 @@ class PandaLoader(RobotLoader):
     ref_posture = "default"
 
 
+class PandaLoaderCollision(RobotLoader):
+    path = "panda_description"
+    urdf_filename = "panda_collision.urdf"
+    urdf_subpath = "urdf"
+    srdf_filename = "panda_collision.srdf"
+    ref_posture = "default"
```
nim65s, line 467:
```suggestion
class PandaLoaderCollision(PandaLoader):
    urdf_filename = "panda_collision.urdf"
    srdf_filename = "panda_collision.srdf"
```

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +459,78 @@ class PandaLoader(RobotLoader):
...
+
+    def __init__(self, verbose=False):
+        super(RobotLoader, self).__init__()
+
+        self.verbose = verbose
```
nim65s, line 472:
```suggestion
        super().__init__(verbose=verbose)
```

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +459,78 @@ class PandaLoader(RobotLoader):
...
+        df_path = join(self.path, self.urdf_subpath, self.urdf_filename)
+        builder = RobotWrapper.BuildFromURDF
+        self.model_path = getModelPath(df_path, self.verbose)
+        self.df_path = join(self.model_path, df_path)
+        self.robot = builder(
+            self.df_path,
+            [join(self.model_path, "../..")],
+        )
```
nim65s, line 480: This is already done in RobotLoader.__init__, why do it again ?

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +459,78 @@ class PandaLoader(RobotLoader):
...
+        cmodel = self.robot.collision_model.copy()
+        list_names_capsules = []
+        # Iterate through geometry objects in the collision model
+        for geom_object in cmodel.geometryObjects:
+            geometry = geom_object.geometry
+            base_name = geom_object.name[:-1]  # Remove the last character from the name
```
nim65s, line 486: why remove that last character ?

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +459,78 @@ class PandaLoader(RobotLoader):
...
+            base_name (str): The base name of the geometry object.
+            existing_names (list): List of names already assigned to capsules.
+
+        Returns:
+            str: Unique capsule name.
+        """
+        suffix = (
+            "capsule_1" if f"{base_name}capsule_0" in existing_names else "capsule_0"
+        )
+        return f"{base_name}{suffix}"
```
nim65s, line 531: This method is not specific to panda, it could go in RobotLoader. Also, `existing_names` should be available in  `self`, so I don't see the pont of making this a staticmethod and then provide it with that context

nim65s, line 531:
```suggestion
        i = 0
        while f"{base_name}_capsule_{i}" in existing_names:
            i += 1
        return f"{base_name}_capsule_{i}"
```

nim65s (changes requested):
I rebased on devel, this might help with CI.

I made a few comments, but also you should add a test

ArthurH91:
Thank you for the review, should be good to go now.

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +474,52 @@ class PandaLoader(RobotLoader):
...
+        super().__init__(verbose=verbose)
+
+        cmodel = self.robot.collision_model.copy()
+        list_names_capsules = []
+        # Iterate through geometry objects in the collision model
+        for geom_object in cmodel.geometryObjects:
+            geometry = geom_object.geometry
+            base_name = geom_object.name[
+                :-2
+            ]  # Remove the last characters from the name
```
nim65s, line 491: Your goal is not to remove the last characters: you are explaining your solution instead of explaining your problem in that comment.

Your goal is to remove the last underscore and the numbers after that. And if those numbers are >9, your solution won't work. Here is a more explicit version:

```suggestion
            base_name = '_'.join(geom_object.name.split('_')[:-1])
```

A regex would be even better, but I think this version is easier to read.
Your comment should include what you explained to me: your goal is to remove what looks like a suffix with an index of the different bodies attached there.

python/example_robot_data/robots_loader.py
```diff
@@ -458,6 +474,52 @@ class PandaLoader(RobotLoader):
...
+                self.robot.collision_model.removeGeometryObject(geom_object.name)
+
+            # Remove spheres associated with links
+            elif isinstance(geometry, hppfcl.Sphere) and "link" in geom_object.name:
+                self.robot.collision_model.removeGeometryObject(geom_object.name)
+
+        # Recreate collision data since the collision pairs changed
+        self.robot.collision_data = self.robot.collision_model.createData()
+
+        self.srdf_path = None
```
nim65s, line 517: This can go before the `__init__` method

python/example_robot_data/robots_loader.py
```diff
@@ -512,6 +528,51 @@ class PandaLoader(RobotLoader):
     ref_posture = "default"
 
 
+class PandaLoaderCollision(PandaLoader):
+    urdf_filename = "panda_collision.urdf"
+    srdf_filename = "panda.srdf"
```
nim65s, line 533:
```suggestion
```

This is already that value in the base class

nim65s (approved):

PR #247, opened 2024-11-08
alex description added

kkouhkil:
Hi there,

I would like to at first thank you for this very useful repository and I would also like to let you know that I am brining in this pull request, the Alex Robot (with three types of end-effector) from Boardwalk Robotics to this repository.


cmastalli (changes requested):
Thanks, @kkouhkil!

Please include a Readme file that reports the source of these files, the license, and any modification with respect to the source. See this example: https://github.com/Gepetto/example-robot-data/tree/master/robots/a1_description

nim65s:
Hi @kkouhkil,
The meshes are 30MiB, could you try to reduce those a bit ? Meshlab and Blender are usually good solutions for this issue.

kkouhkil:
> Hi @kkouhkil, The meshes are 30MiB, could you try to reduce those a bit ? Meshlab and Blender are usually good solutions for this issue.

Hi there, thank you for your text. I did bring the size down to 19MiB, I would be appreciated to know if it is good enough as each mesh file is now under 1MiB.

cmastalli (approved):

nim65s:
This is still big :/

I had a look at the meshes, and I think there is no reason to be that big. And then I found that these are plain ASCII files. I should have seen it earlier, sorry.

With https://github.com/Neizvestnyj/stl-to-obj it seems easy to go down to 11M. Can you do that, and check everything is still OK for your use case ?

nim65s:
I'll handle it and make a release I think

nim65s:
ref. #254

nim65s:
I don't have time for now, so let's merge it. I'll deal with the excessive size in another way.

PR #252, opened 2024-11-26
Improvements to Underwater Vehicle Manipulators

CarlosSuarezZapico:
Hi, I am making this pull request for the following improvements in the underwater vehicle manipulator:

    1. Rotating the Bluevolta in roll and yaw -pi and pi, respectively to have a good initial orientation. 
    2. Define a reasonable arm posture in Bluevolta + Bravo7
    3. Substitute negative values for the joint axis, i.e., change this <axis xyz="0 0 -1"/> by <axis xyz="0 0 1"/> in the URDF. This triggers issues with Pinocchio

### DIFF
--- a/robots/bluevolta_description/srdf/bluevolta_bravo7_gripper.srdf
+++ b/robots/bluevolta_description/srdf/bluevolta_bravo7_gripper.srdf
@@ -18,12 +18,12 @@
 
     <!--GROUP STATES: Purpose: Define a named state for a particular group, in terms of joint values. This is useful to define states like 'folded arms'-->
     <group_state name="standing" group="whole_body">
-        <joint name="joint1" value="0.0"/>
-        <joint name="joint2" value="2.305"/>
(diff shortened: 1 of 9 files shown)


nim65s (approved):
These are breaking changes, so I would usually try to avoid those, but since I did not manage to make a release with those robots, I think this is fine.

cmastalli (approved):
Thanks for improving the URDFs and SRDFs, @CarlosSuarezZapico!

Thanks for reviewing this PR, @nim65s!

Just a bit of context, I asked Carlos to change the URDFs to Pinocchio doesn't like negative signs for defining the rotation axis. However, the model remains the same. Additionally, I requested Carlos to change the nominal posture since it was impractical.

These URDFs and SRDFs are still experimental as we had to develop them ourselves. The company doesn't provide them. We are unsure about the inertial tensors and, in the future, we plan to run an estimation procedure. We likely have to create another PR with changes in the inertial properties, something we'll do in the next months.

PR #260, opened 2025-01-10
Update Go2 collisions 

EtienneAr:
Hello,

The Go2 collisions were very minimalistic, so I updated it to be more accurate. (See picture below)

I used convex hulls of visual meshes to do so.

I split the calf body in 2 convex shapes to keep a bit of the curve, and I removed the collision mesh of the front hips (as it was almost always in the body, so not very useful).

I also updated the srdf to exclude the appropriate pairs of collisions.

Before: 
Screenshot from 2025-01-08 15-14-51

After:
Screenshot from 2025-01-10 13-26-21

(I was tempted to split the body in several smaller convex mesh, and the thigh as well, for a more accurate representation of the collisions. But finally I thought keeping them as just one hull in not that big of an approximation and significantly reduces the number of collision pairs, which is always nice. This choice can be discussed)

Cheers


robots/go2_description/urdf/go2.urdf
```diff
@@ -36,88 +32,11 @@ Stephen Brawner (brawner@gmail.com)
...
     </visual>
     <collision>
-      <origin rpy="0 0 0" xyz="0 0 0" />
-      <geometry>
-        <box size="0.3762 0.0935 0.114" />
-      </geometry>
-    </collision>
-  </link>
-  <link
-    name="Head_upper">
```
nim65s, line 46: I hope this link was not used by anybody for anything else that collisions.
Technically, removing this is a breaking change, but I think we can take the risk.

nim65s (commented):
This really looks excellent work, thanks Étienne !

Can you add a changelog entry, including the list of the links you removed ?
The probability that somebody needs that is very low, but not zero, so I think this should at least be properly documented.

robots/go2_description/urdf/go2.urdf
EtienneAr: Yes, you are completely right. I thought that since the Go2 model was added recently by @edantec it was still early enough to do this kind of changes...

nim65s (approved):

cmake
cmastalli, line 1: Is there any motivation behind updating the cmake submodule?

EtienneAr: Oops, no, sorry. It must have accidentally happened during a git add...

cmastalli: No worries! could you revert it?

EtienneAr: Yes, sure

PR #261, opened 2025-01-13
Fix ROS2 installation

EtienneAr:
Hello,

Currently, when working with ros2 (ament /colcon), users usually need to set environment variables by hand in order for some tools (such has RViz 2) to find **example-robot-data** packages. (for instance, the `AMENT_PREFIX_PATH` variable)

Thanks to @olivier-stasse PR, https://github.com/jrl-umi3218/jrl-cmakemodules/pull/717, the **jrl-cmakemodules** can create the necessary files during installation so that when the user `source install/setup.bash` the env variable are set.

So in this PR I propose to update the **jrl-cmakemodules** submodule (and add the small colcon.pkg file)

### DIFF
--- a/colcon.pkg
+++ b/colcon.pkg
@@ -1,0 +1,6 @@
+{
+    "hooks": [
+        "share/example-robot-data/hook/ament_prefix_path.dsv",
+        "share/example-robot-data/hook/python_path.dsv"
+    ]
+}


EtienneAr:
To activate this feature from the jrl-cmakemodules, the user need to set the flag `-DBUILDING_ROS2_PACKAGE=ON`.
However in [pinocchio](https://github.com/stack-of-tasks/pinocchio) this feature is always on (see [CMakeLists.txt#L526-L537](https://github.com/stack-of-tasks/pinocchio/blob/2e86eca560beddb178c30eb9d0174a51cd76a7ae/CMakeLists.txt#L526-L537))

Should we activate it by default here aswell ?

PS: pinocchio does not use the jrl-cmakemodules feature to do it, but since one is a copy/paste of the other, I'll do a PR to remove duplicate code

nim65s:
HI @EtienneAr,

Thanks for this PR, this has been on my todolist for a while.
I think we should activate this feature by default in jrl-cmakemodules. I'll open a PR for that.

EtienneAr:
Thank you @nim65s

nim65s (approved):

cmastalli (approved):

PR #278, opened 2025-04-04
ur_description: remove all "LightGrey" `<material>` tags from URDF files

ManifoldFR:
This triggered Pinocchio's URDF parser to set `geometry_object.overrideMaterial` to `true` on all loaded `GeometryObject`s, although the loaded meshes have perfectly serviceable materials.

cc @nim65s

### DIFF
--- a/robots/ur_description/urdf/ur10_joint_limited_robot.urdf
+++ b/robots/ur_description/urdf/ur10_joint_limited_robot.urdf
@@ -34,9 +34,6 @@
       <geometry>
         <mesh filename="package://example-robot-data/robots/ur_description/meshes/ur10/visual/base.dae"/>
       </geometry>
-      <material name="LightGrey">
-        <color rgba="0.7 0.7 0.7 1.0"/>
-      </material>
     </visual>
     <collision>
       <geometry>
@@ -62,9 +59,6 @@
       <geometry>
         <mesh filename="package://example-robot-data/robots/ur_description/meshes/ur10/visual/shoulder.dae"/>
       </geometry>
-      <material name="LightGrey">
-        <color rgba="0.7 0.7 0.7 1.0"/>
-      </material>
     </visual>
     <collision>
       <geometry>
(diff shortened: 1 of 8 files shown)


nim65s (approved):
Thanks !

cmastalli:
@ManifoldFR -- have you checked that Rviz is working properly? I think Rviz might parse or not Collada files. This depends on how they have been generated.

ManifoldFR:
I do not use RViz. But @nim65s told me he would look at the result inside other viewers before merging.

What I know is that materials for Collada files only show up correctly in the Meshcat visualizer because it is broken and an internal bug somehow compensates for it. I think the material override is not properly applied because of a bug in Meshcat's Collada loader (or at least, the one we call through the [`DaeMeshGeometry` class we had to implement ourselves](https://github.com/stack-of-tasks/pinocchio/blob/1a8c32eb03c10b536d9de93d7b3bf9791e9e434b/bindings/python/pinocchio/visualize/meshcat_visualizer.py#L116-L197) because Collada support is bugged in the _released_ PyPI version of Meshcat). meshcat-python itself is kind of a complete mess and our visualizer kind of stacks weird patches on top of it.

If RViz doesn't know how to properly load materials from Collada and we have to 'hack' our URDF files to force-load a bogus material just for it - and break the workflow for viewers that _aren't_ themselves broken, then I'm not sure what to do lol

cmastalli:
> I do not use RViz. But @nim65s told me he would look at the result inside other viewers before merging.
> 
> What I know is that materials for Collada files only show up correctly in the Meshcat visualizer because it is broken and an internal bug somehow compensates for it. I think the material override is not properly applied because of a bug in Meshcat's Collada loader (or at least, the one we call through the [`DaeMeshGeometry` class we had to implement ourselves](https://github.com/stack-of-tasks/pinocchio/blob/1a8c32eb03c10b536d9de93d7b3bf9791e9e434b/bindings/python/pinocchio/visualize/meshcat_visualizer.py#L116-L197) because Collada support is bugged in the _released_ PyPI version of Meshcat). meshcat-python itself is kind of a complete mess and our visualizer stacks some weird patches on top of it.

Thanks for the explanation, @ManifoldFR !

> If RViz doesn't know how to properly load materials from Collada and we have to 'hack' our URDF files to force-load a bogus material just for it, and break the workflow for viewers that _aren't_ themselves broken, then I'm not sure what to do lol

I hope we do not have to worry about this. We can discuss this further if needed.

PR #303, opened 2025-06-26
Rescaling of human model based on anthropometry

MaximeSabbah:
Allows to load the human model specifying height, weight and gender. The model automatically scales based on this.


python/example_robot_data/robots_loader.py
```diff
@@ -753,6 +779,1242 @@ def __init__(self, verbose: bool = False):
...
+                    )
+                    * dict_sgmt["mass"],
+                    5,
+                )
+                dicts.append(dict_sgmt)
+            elif name == "middle_head":
+                length = (0.1308 if gender == "f" else 0.1310) * height
+                sgt_lengths["middle_head_0"] = length
+                dict_sgmt["mass"] = np.round(
+                    (0.067 if gender == "f" else 0.067) * weight, 2
```
nim65s: Something is odd here

python/example_robot_data/robots_loader.py
```diff
@@ -823,12 +2085,27 @@ def __init__(self, verbose: bool = False):
...
+            height = kwargs.pop("height")
+            weight = kwargs.pop("weight")
+            gender = kwargs.pop("gender")
+        else:
+            height = weight = None  # mark for fallback
+            gender = "m"  # default
+
+        inst = HumanLoader(height=height, weight=weight, gender=gender, verbose=verbose)
+    else:
+        inst = ROBOTS[name](verbose=verbose)
```
nim65s: I'm not fan of this logic with added complexity in the shortcut. I think for this use case, the user should just use
```python
robot = HumanLoader(height=…)
```

Do you really need that ?

python/example_robot_data/robots_loader.py
```diff
@@ -740,10 +741,35 @@ class HumanLoader(RobotLoader):
         ]
     )
 
-    def __init__(self, verbose: bool = False):
+    def __init__(
```
nim65s: that is a big piece of code. Maybe we should start splitting this file into multiple ones. But maybe this can be done in another PR.

nim65s (commented):
I rewrote the history a bit, cleaned a detail or two, and have a couple comments:

python/example_robot_data/robots_loader.py
nim65s: (CI is failing because of this)

python/example_robot_data/robots_loader.py
MaximeSabbah: Not particularly, I was not sure how to do it, so I did both ways. I agree that we can make things easier here

PR #330, opened 2025-10-15
ROS: update

nim65s:


package.xml
```diff
@@ -12,13 +12,10 @@
 
   <url type="repository">https://github.com/gepetto/example-robot-data</url>
 
-  <!-- The following tags are recommended by REP-136 -->
-  <exec_depend condition="$ROS_VERSION == 1">catkin</exec_depend>
   <exec_depend condition="$ROS_VERSION == 2">ament_cmake</exec_depend>
   <build_depend>git</build_depend>
-  <depend condition="$ROS_PYTHON_VERSION == 2">python</depend>
+  <build_depend>jrl_cmakemodules</build_depend>
```
cmastalli, line 17: Does this support ROS1? (I know it's deprecated but still lot of people are using ROS1)

nim65s: It is not deprecated, but end-of-life.
And we are not maintaining it: we don't have CI for it.

In any case, I don't think this PR will change anything for users still on ROS 1.

PR #334, opened 2025-10-25
Import SO100/101 arms

kaidokert:
Importing from https://github.com/TheRobotStudio/SO-ARM100.git

Tested in Pinocchio with simple loaders.


kaidokert:
Fwiw, i also have a simplified collision geometry version for this, with about a dozen collision boxes - shaped so that it only prevents self-collision.

nim65s:
Hi,

I've rebased on devel, squashed the commits, added unit tests, and remove the *.part files that seem unecessary

image

image


@kaidokert : thanks for this contribution. Can we merge this, or do you need the *.part ?

kaidokert:
No need for the "*.parts" - it was just there and i copied it wholesale from upstream.

PR #338, opened 2025-11-03
Installing example-robot-data without hppfcl

cmastalli:
This PR enables us to install this package with the minimal Pinocchio configuration, i.e., without hppfcl.


python/example_robot_data/panda.py
```diff
@@ -1,9 +1,13 @@
 from os.path import join
 
-import hppfcl
 import numpy as np
 import pinocchio as pin
 
+try:
+    import hppfcl as fcl
+except Exception:
```
nim65s:
```suggestion
except ImportError:
```

nim65s (approved):

nim65s (approved):
thanks :)

PR #354, opened 2026-01-20
Add Unitree G1 robot description 

jose-rojasr:
This PR adds the G1 humanoid robot model to example-robot-data.


nim65s (changes requested):
Hi,

Thanks for this contribution.

Could you add a unit test ?

Also, many STL files are too heavy for this repository. I guess all files above 1M can be decimated without having a huge visual impact. Some can be decimated multiple times. You can for example use meshlab for that

jose-rojasr:
> Hi,
> 
> Thanks for this contribution.
> 
> Could you add a unit test ?
> 
> Also, many STL files are too heavy for this repository. I guess all files above 1M can be decimated without having a huge visual impact. Some can be decimated multiple times. You can for example use meshlab for that

Thanks for the suggestions.

I added a unit test as part of the standard unittest/test_load.py.

For mesh decimation, I processed all STL files in robots/g1_description/meshes and decimated every mesh larger than 1 MB, targeting just under 1 MB per file to preserve as much visual quality as possible.

nim65s (approved):

PR #346, opened 2025-12-03
Added Centauro to example-robot-data robots

AndrePatri:
The new centauro_description weights 9.3 MB total, all meshes are below 0.3MB.

PR #361, opened 2026-02-24
Added colors to Centauro's URDFs

AndrePatri:
As requested in [this](https://github.com/Gepetto/example-robot-data/pull/346) previous pull request, I've added realistic colors to Centauro. Now it should appear like this:
Screenshot from 2026-02-24 11-14-06
Additionally, I've added back the original visual for the pelvis link, instead of using the simplified collision one. The new mesh weights 1.8 MB.


cmastalli (commented):
Thanks for improving the Centauro's visualization!

Would it be possible to reduce the size of the pelvis mesh?

AndrePatri:
I've reduced the pelvis mesh size to 0.88 MB approximately.  I can reduce it a bit more,  but it'll look quite ugly. All the other meshes are below 0.3MB, worst case

nim65s (approved):
Thanks !

I rebased & squashed the git history, to avoid the big files from earlier to end up in the repo forever for nothing.

cmastalli (approved):

PR #369, opened 2026-03-30
Tiago pro add srdf

clementPene:
Added a srdf for Tiago Pro with one gripper per arm 
Updated robot Loader


nim65s:
What do you mean by "HPP gripper" ?

PR #374, opened 2026-06-05
Use coal instead of hppfcl

jorisv:
This is breaking change.

@nim65s Since python package is importing coal directly, should we add coal in package.xml dependecies ?

### DIFF
--- a/python/example_robot_data/utils.py
+++ b/python/example_robot_data/utils.py
@@ -130,7 +130,7 @@ class RobotLoader:
                 self.ref_posture,
             )
 
-            if pin.WITH_HPP_FCL and pin.WITH_HPP_FCL_BINDINGS:
+            if pin.WITH_COLLISION:
                 # Add all collision pairs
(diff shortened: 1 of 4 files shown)


python/example_robot_data/utils.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, verbose=False):
                 self.ref_posture,
             )
 
-            if pin.WITH_HPP_FCL and pin.WITH_HPP_FCL_BINDINGS:
+            if pin.WITH_COLLISION:
```
nim65s, line 133: This add a `pinocchio >= 4` constraint, but I think this is fine

nim65s (commented):
> should we add coal in package.xml dependecies ?

yes, please

nim65s (approved):

nim65s:
@mergifyio rebase

nim65s:
@mergifyio rebase

PR #383, opened 2026-08-18
Add B1 and Kangaroo closed-loop models

Sergim96:
## Summary

This PR adds several closed-loop robot models.

### Unitree B1

- `b1_leg`: serial three-joint leg without constraints.
- `b1_leg_3D`: leg with a position-level four-bar closure.
- `b1_leg_6D`: leg with a pose-level four-bar closure.
- `b1_closed_loop`: complete floating-base B1 with closed-loop four-bar mechanisms on all four legs.

### Kangaroo

- `kangaroo`: serial model without loop constraints.
- `kangaroo_knee_kin`: knee mechanisms closed on both legs.
- `kangaroo_knee_1ankle_kin`: knee and first ankle linkage stage.
- `kangaroo_knee_2ankle_kin`: knee and both ankle linkage stages.
- `kangaroo_legs_kin`: complete leg mechanisms, including knee, ankle, and hip actuator loops.

## SRDF constraint utility

This PR also adds a utility that parses `<loop_constraint>` entries from SRDF files and resolves each frame pair into its parent joints, placements, and active constraint DoFs.

When a constraint provides an SRDF `mask`, the utility uses it directly. When the mask is absent, it numerically infers the active DoFs by perturbing the joints supporting the constraint frames and measuring their linear and angular motion. Currently, `b1_leg_6D` uses this numerical fallback, while the remaining closed-loop models provide explicit masks.

The models assume that the SRDF reference configuration `q0` already satisfies all loop constraints.


nim65s:
Thanks, but this repo is already huge.
Could you check if this could go to https://github.com/Gepetto/example-parallel-robots instead ? we already have a few closed-loop models there

cmastalli:
> Thanks, but this repo is already huge. Could you check if this could go to https://github.com/Gepetto/example-parallel-robots instead ? we already have a few closed-loop models there

I believe we should start supporting closed-loop kinematics in example-robot-data. This will allow us to continuously test new features introduced in Crocoddyl (see this PR https://github.com/loco-3d/crocoddyl/pull/1525) without depending on another upstream package.

What do you think?

nim65s:
hum, yes, that would make sense. I'll check a few things and come back to this PR asap

PR #387, opened 2026-09-02
Fix trunk segment ratios and trunk visual scaling in the human model

MaximeSabbah:
Several related problems left the trunk wrong: a torso sitting below the thoracic joint with the head floating detached above it, worse the taller the subject, and a pelvis narrower than the torso above it.

Trunk segment lengths. The joints of this model are the intersegmental centres L5/S1, T12/L1 and C7/T1, following Dumas et al., as the limb ratios already do. The abdomen and thorax lengths, however, came from De Leva's trunk sub-segments, which are bounded by the xiphion and the omphalion instead. Those are different boundaries, so the thoracolumbar joint sat about 7 cm too high for a 1.7 m subject. Rescaling the same model from motion capture then moved that joint back down, which is what detached the head from the torso. Use ratios consistent with the joints the model actually has, keeping the trunk total, L5/S1 to C7/T1, unchanged at 0.2497 H (f) and 0.2627 H (m). A marker-based calibration now shifts the joint by under 2 mm instead of 71 mm.

Pelvis width. The inter-ASIS ratios had the sexes the other way round, leaving men with a narrower pelvis than women of the same stature. Use the inter-ASIS distances Shankar et al. measured with calipers on 100 adults: 0.1368 H for women and 0.1286 H for men.

Torso mesh. torso_mesh.STL measures 52.0 along its Y axis, not the 35 it was divided by. It also stands for the whole trunk rather than the thorax segment alone, so its length now spans from the lumbar joint to the cervical joint; sized to the thorax segment it stops short of the pelvis. Its width is taken from the pelvis, so the two read as one body instead of a torso overhanging the hips.

Torso visual offset. The mesh does not begin at its link origin, so the URDF offsets it. That offset was both the wrong length and fixed, while the mesh is rescaled for every subject, so it could only suit one size. Correct it and rescale it with the mesh. The rescaling is applied to every visual, so any mesh carrying an offset now keeps its place. The URDF scales are updated to match, so the unscaled model is consistent with a scaled one.

Verified from 1.55 m to 1.87 m: the torso spans the trunk with its base on the lumbar joint and its top on the cervical joint, both within 4 mm, the neck emerges from the top of the chest, and torso and pelvis share one width at exactly the inter-ASIS ratio.

Visuals and segment lengths only: no inertia or joint axis is touched.


nim65s (approved):
this is fine for me, but if these details are really important, we should probably have a separate package to properly host them.

MaximeSabbah:
It is mainly visual, I think it is not required to host it somewhere. It is also normally commented a bit in the code.

PR #388, opened 2026-09-23
ROS: rename package to example_robot_descriptions

nim65s:
Hi,

I tried to add this projet to ROS buildfarm, so that tsid/crocoddyl/HPP/mim_solvers/ndcurves/etc. can more easily use `<test_depend>example_robot_data</test_depend>` in their `package.xml`.

ROS PMC is OK with that, but believe the name would not be suitable for ROS, and did propose `example_robot_descriptions` instead.

I think this name is good, and I don't think changing **only the ROS name** is too big of an issue.

This commit is doing that, only in our `package.xml`, so this will have to be adaptated in the `package.xml` of dependent packages. On this topic, *this is a breaking change*.

cc. @cmastalli & @MaximilienNaveau about this.

Just to be clear, I don't intend to rename github name/url, cmake project, c++ header dir, python module, and packages in AUR, conda, PyPI, nixpkgs, etc. at least until we get a lot of feedback and a very good consensus.

ref. https://github.com/ros/rosdistro/pull/53263
ref. https://github.com/Gepetto/example-robot-data/issues/1


cmastalli:
Hi @nim65s 

In my opinion, the name example-robot-data is quite misleading. This is why I support replacing it with example-robot-descriptions broadly (i.e., github name/url, cmake project, C++ header, etc.). I'm not sure what the right transition action is, but it seems a justified action.

nim65s:
Ok, thanks for your opinion, I'll keep discuss this with more people, and if nobody raise more objections I'll start working on a bigger migration plan

